PR #4, opened 2014-11-25
Changed most of the plain-text blocks to a use a different font

rgaudin:
- Added a @sans-serif-family var.
- Kept number-based texts on serif/times

screen shot 2014-11-25 at 4 20 29 pm

@nicpottier Just so you can easily preview/test what it would look like with a different font.


nicpottier:
Closing as the designer said clashing fonts are cool. :)

PR #11, opened 2015-01-09
Jobspage

xkmato:
This is a the jobs page(requested by Nigeria). A page filled with many rss and social network boxes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,9 @@ db/
 
 # symlink to dash for dev
 dash
+
+#collected by pip
+src
+
+#some play around files
+ureport/playground
--- a/pip-freeze.txt
+++ b/pip-freeze.txt
@@ -34,6 +34,7 @@ redis==2.10.1
 requests==2.3.0
 simplejson==3.6.0
 six==1.7.3
+feedparser==5.1.3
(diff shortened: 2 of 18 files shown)


ewheeler:
Cool! Would it be possible to add a configuration setting for this (maybe one that can be toggled by an administrator)? I'm not sure that every U-report country will be interested in having a jobs page at the moment and we don't want them to have a sad, empty jobs page. We should let them activate the feature once they're prepared to support it.

xkmato:
Yep... I see what you mean. Right now it is configurable in settings.py but still not by the site Administrators(actual site users). Will look into adding that

nicpottier:
@norkans7 Can you work on pulling this in? (obviously will need a merge from master as well)

Couple things we'll need to do:
 * Jobs et all need to have an Org id on them that is auto populated from the current site when administering
 * We'll need an Org config option for whether Jobs are shown for a site or not
 * Clean up the templates, note we'll need these to be mobile compatible with the new internet.org styles, we can chat about how to test this tomorrow
 * Clean up styles to use similar styles as we have currently
 * Add tests so coverage stays at 100%

norkans7:
replaced by #16

PR #13, opened 2015-03-04
delete poll questions associated with old rulesets

norkans7:


nicpottier:
Great, thanks Norbert.

PR #16, opened 2015-03-20
Xkmato jobspage

norkans7:


norkans7:
Please review

templates/public/jobs.haml
```diff
@@ -0,0 +1,165 @@
...
+            .hero-small-background
+            .row
+              .col-md-7
+                .hero-dark-content
+                  .hero-title.hero-top-title.primary-bg-color
+                    - trans "JOBS"
+                  .hero-description
+                    {{ d_block.content }}
+                  .hero-info
+                    - trans "Scroll to see job details"
```
nicpottier: Maybe instead of this being constant this should be the summary for the block?

ureport/jobs/models.py
```diff
@@ -0,0 +1,52 @@
...
+    org = models.ForeignKey(Org,
+                            help_text=_("The organization this job source is for"))
+
+    def __unicode__(self):
+        return self.title
+
+    def get_entries(self):
+        entries = []
+        try:
+            feed = feedparser.parse(self.source_url)
```
nicpottier: We should cache these so we don't fetch them on every request, perhaps once a day?

nicpottier: Better yet, is there a JS library we could use to have their browser do the fetching?

ureport/jobs/views.py
```diff
@@ -0,0 +1,37 @@
...
+        fields = ('is_featured', 'title', 'source_type', 'source_url', 'widget_id')
+
+        def pre_save(self, obj):
+            obj = super(JobSourceCRUDL.Create, self).pre_save(obj)
+            obj.org = self.request.org
+            return obj
+
+    class Update(OrgObjPermsMixin, SmartUpdateView):
+        success_url = '@jobs.jobsource_list'
+        success_message = _("Your job source has been added successfully")
```
nicpottier: updated not added.

ureport/settings_common.py
```diff
@@ -185,7 +185,8 @@
                      dict(name='state_label', field=dict(help_text=_("The label of the Contact Field that contains the State of reporters")), superuser_only=True),
                      dict(name='district_label', field=dict(help_text=_("The label of the Contact Field that contains the District of reporters")), superuser_only=True),
                      dict(name='male_label', field=dict(help_text=_("The label assigned to U-reporters that are Male.")), superuser_only=True),
-                     dict(name='female_label', field=dict(help_text=_("The label assigned to U-reporters that are Female.")), superuser_only=True)]
+                     dict(name='female_label', field=dict(help_text=_("The label assigned to U-reporters that are Female.")), superuser_only=True),
+                     dict(name='show_jobs', field=dict(help_text=_("If the jobs page should be shown on the public site"), required=False))]
 #                     dict(name='featured_state', field=dict(help_text=_("Choose the featured State of reporters shown on the home page")))]
```
nicpottier: So what are the valid values here? "True" and "False"? Maybe we should say so in the help?

ureport/tests.py
```diff
@@ -129,6 +131,33 @@ def create_org(self, subdomain, user):
         self.assertEquals(Org.objects.filter(subdomain=subdomain).count(), 1)
         return Org.objects.get(subdomain=subdomain)
 
+class UreportJobsTest(DashTest):
+    FB_SOURCE = 'http://www.facebook.com/%s'
```
nicpottier: This doesn't below in ureport/tests.py

ureport/settings_common.py
norkans7: here we use a checkbox https://github.com/rapidpro/dash/pull/16

or we need to change this config to tell the field class to use(CharField, BooleanField,...)

ureport/tests.py
norkans7: I wanted to reuse the methods both in the public/tests and jobs/tests

templates/public/jobs.haml
norkans7: I had the block_type with has_summary = False since we do not have a way to configure the content.

so instead of having the use to think which one to use? I disabled the summary use the content which is always shown

ureport/jobs/views.py
norkans7: I change this

templates/public/jobs.haml
nicpottier: Wouldn't they configure the content the same way they do for any other quickblock?

norkans7: Yes it is the same way but I do not want to show both the summary and the content textarea, I just show the content

norkans7: https://github.com/rapidpro/ureport/blob/xkmato-jobspage/ureport/jobs/migrations/0002_auto_20150320_1404.py#L18

nicpottier: Ah ok, didn't realize we weren't showing the content anywhere so this takes the place of that.

ureport/tests.py
nicpottier: Ok.

ureport/jobs/models.py
norkans7: I used the cache,  so far I am failing to use jquery to fetch xml response 

error: No 'Access-Control-Allow-Origin' header is present on the requested resource.

ureport/jobs/migrations/0002_auto_20150320_1404.py
```diff
@@ -0,0 +1,44 @@
...
+
+def generate_job_block_types(apps, schema_editor):
+    User = apps.get_model("auth", "User")
+    root = User.objects.filter(username="root").first()
+
+    if not root:
+        root = User.objects.filter(username="root2").first()
+
+    if not root:
+        root = User.objects.create(username="root2")
```
nicpottier: Why do we have this? Why not just create root if it doesn't exist?

norkans7: So that we still allow creating the superuser root and give it a password

norkans7: I changed it,though we have that in previous migrations for the previous reason

nicpottier:
Nice, thanks Norbert.

PR #19, opened 2015-04-01
Valid countries

norkans7:


nicpottier:
Looks good norbert.

norkans7:
This is ready for review @nicpottier

ureport/countries/views.py
```diff
@@ -0,0 +1,15 @@
+import pycountry
+from smartmin.views import SmartCRUDL, SmartListView
+from .models import CountryAlias
+
+class CountryALiasCRUDL(SmartCRUDL):
```
nicpottier: Typo with case in ALias.

nicpottier:
Looks good Norbert, were you able to find a projection that had slightly less focus on Greenland? Would be nice to have that as I'm sure it will come up.

ureport/countries/views.py
norkans7: Okay I fix that

norkans7:
for the projection I used http://jsfiddle.net/gh/get/jquery/1.11.0/highslide-software/highcharts.com/tree/master/samples/mapdata/custom/world-eckert3

nicpottier:
Nice, perfect.

PR #21, opened 2015-04-15
Dynamic landing page

norkans7:


nicpottier:
This looks good apart from my comments about having the flags be part of U-Report instead of dash.

PR #22, opened 2015-04-16
Migrate org backgrounds

norkans7:


ureport/assets/models.py
```diff
@@ -0,0 +1,19 @@
...
+class Background(SmartModel):
+    org = models.ForeignKey(Org, verbose_name=_("Org"), related_name="ureport_backgrounds",
+                            help_text=_("The organization in which the image will be used"))
+
+    name = models.CharField(verbose_name=_("Name"), max_length=128,
+                            help_text=_("The name to describe this background"))
+
+    background_type = models.CharField(max_length=1, choices=BACKGROUND_TYPES, default='P', verbose_name=_("Background type"))
+
+    image = models.ImageField(upload_to='org_bgs', help_text=_("The image file"))
```
nicpottier: I think we need ```from __future__ import unicode_literals``` at the top of this, which is why in the migiration it is doing ```b"blah blah"``` instead of ```u"blah blah"```

nicpottier: Also, since only one of these is actually a background, maybe we should call these something else?

OrgImage?

nicpottier: Seems like the app being named ```assets``` is a bit confusing given the current naming too unless we are stuffing more things into ```assets```

ureport/assets/migrations/0001_initial.py
```diff
@@ -0,0 +1,35 @@
...
+
+from django.db import models, migrations
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('orgs', '0008_org_timezone'),
```
nicpottier: Oooh.. swappable_dependency!

PR #24, opened 2015-04-28
Add s3 support for populate image migrations

norkans7:


ureport/assets/migrations/0002_populate_images.py
```diff
@@ -10,22 +11,29 @@ def populate_images(apps, schema_editor):
...
+
+                image_obj = Image()
+                image_obj.org = org_bg.org
+                image_obj.name = org_bg.name
+                image_obj.image_type = org_bg.background_type
+                image_obj.is_active = org_bg.is_active
+                image_obj.created_by = org_bg.created_by
+                image_obj.modified_by = org_bg.modified_by
+                image_obj.image.save(image_filename, django_image_file, save=True)
+
```
nicpottier: Why is this in a finally vs being below the except NotImplementedError above?

norkans7: none I change it

nicpottier:
Yep.

PR #29, opened 2015-05-05
Redirect chooser to global site

norkans7:


ureport/public/context_processors.py
```diff
@@ -9,4 +13,22 @@ def set_is_iorg(request):
         is_iorg = True
 
     print "IS_IORG (%s): %s" % (request.META.get('HTTP_VIA'), is_iorg)
```
nicpottier: We can probably get rid of this print. (I know I added it)

ureport/public/context_processors.py
```diff
@@ -9,4 +13,22 @@ def set_is_iorg(request):
...
-    return dict(is_iorg=is_iorg)
\ No newline at end of file
+    return dict(is_iorg=is_iorg)
+
+def linked_sites(request):
```
nicpottier: I don't think we need this on every page, only when showing a global site no? Maybe it should really be a template tag instead?

ureport/public/urls.py
```diff
@@ -3,10 +3,10 @@
 from django.views.decorators.cache import cache_page
 from .views import IndexView, NewsView, AboutView, PollsView, PollReadView, PollQuestionResultsView, JobsView
 from .views import BoundaryView, UreportersView, StoriesView, StoryReadView, JoinEngageView, ReportersResultsView
-from .views import CountriesView, Chooser
+from .views import CountriesView, chooser
 
 urlpatterns = patterns('',
-    (r'^home/$', Chooser.as_view(), {}, 'public.home'),
+    (r'^home/$', chooser, {}, 'public.home'),
```
nicpottier: So aren't we going to have two levels of redirection here? One for the middleware redirection to this view, then this view redirection to 'www'.

Would be nice to simplify that, or even just create a Global view that extends our current public_index for an org that we redirect to instead.

norkans7: If I extend the Index view for the particular org I can put the linked org in its context and not use the template tag?

ureport/polls/templatetags/ureport.py
```diff
@@ -168,3 +172,8 @@ def render(self, context):
 
 # register our tag
 lessblock = register.tag(lessblock)
+
+@register.inclusion_tag('public/org_flags.haml')
```
nicpottier: Does this work? with .haml?

norkans7: It works for both, I change that to html and fix variables

ureport/utils/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    all_orgs = Org.objects.filter(is_active=True).order_by('name')
+
+    linked_sites = list(getattr(settings, 'PREVIOUS_ORG_SITES', []))
+
+    # populate a ureport site for each org so we can link off to them
+    for org in all_orgs:
+        host = settings.SITE_HOST_PATTERN % org.subdomain
+        org.host = host
+        if org.get_config('is_on_landing_page'):
+            flag = Image.objects.filter(org=org, is_active=True, image_type='F').first()
```
nicpottier: Use the constant here instead of 'F'

nicpottier:
Shipit.

PR #44, opened 2015-06-25
Fetch poll results right after a poll in created or updated

norkans7:
Fire tasks to fetch data right after a poll is added

@nicpottier plz review


nicpottier:
Shipit!

PR #45, opened 2015-07-13
allow custom domains

norkans7:


nicpottier:
Will this change the links from the org_chooser as well to be the full domain versions?

norkans7:
Ohh I had missed that but that is a change in Dash, I add that

nicpottier:
Yep.

PR #46, opened 2015-07-13
fix hostname setting and typos

norkans7:


nicpottier:
Looks good. You should add a .travis file for UReport so we can get automated build notifications here as well.

PR #49, opened 2015-07-15
Global tweaks

nicpottier:
- Changes wording on banner text
- Changes U-REPORTER STORY to STORY
- Hide U-Reporter stats for Global org
- Also update our translation files.


templates/public/index.haml
```diff
@@ -129,81 +130,81 @@
...
+      .level-1-title-description.secondary-color
+      .card.card-padding
+        .row
           .col-md-4
             .hr-header.primary-color
-              -trans "RATE BY AGE"
-
-            .age-stats
-              -include "public/mobile_age_include.html"
+              -if org|config:"is_global"
```
norkans7: We  can remove this if block is no longer relevant

nicpottier: Which block do you mean?

norkans7: -if org|config:"is_global" to just show the title as "@ ureportglobal Members" down

nicpottier:
Ok @norkans7 ?

norkans7:
Yep

PR #50, opened 2015-07-15
tweak hostname settings for tests

norkans7:


ureport/settings.py.postgres
```diff
@@ -14,7 +14,8 @@ API_RESULT_CACHE_TIME = 60 * 60 * 6
 
 SITE_API_HOST = 'rapidpro.io'
 API_ENDPOINT = 'https://api.rapidpro.io'
-HOSTNAME = 'ureport.io'
+HOSTNAME = 'ureport.io' if TESTING else 'ureport.io:8000'
```
nicpottier: Should this be ```localhost:8000``` or do you just map ureport.io to localhost on your box?

norkans7: yep I change that

PR #51, opened 2015-07-22
strip extra quotes and spaces on country text for countries validatio…

norkans7:
…n webhook

### DIFF
--- a/ureport/countries/migrations/0002_auto_20150722_1524.py
+++ b/ureport/countries/migrations/0002_auto_20150722_1524.py
@@ -5,17 +5,17 @@ import re
 from django.db import models, migrations
 
 
-def name_stemming(name):
+def normalize_name(name):
     words = re.split(r'\W+', name, re.UNICODE)
     return " ".join([word.lower() for word in words if word])
 
 
-def populated_stemmed_name(apps, schema_editor):
+def normalize_country_aliases(apps, schema_editor):
     CountryAlias = apps.get_model("countries", "CountryAlias")
 
     for alias in CountryAlias.objects.all():
         name = alias.name
-        new_name = name_stemming(name)
+        new_name = normalize_name(name)
         alias.name = new_name
         alias.save()
 
@@ -27,5 +27,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(populated_stemmed_name),
+        migrations.RunPython(normalize_country_aliases),
     ]
--- a/ureport/countries/migrations/0002_auto_20150722_1524.py
+++ b/ureport/countries/migrations/0002_auto_20150722_1524.py
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-import re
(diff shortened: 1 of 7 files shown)


ureport/public/views.py
```diff
@@ -327,6 +327,7 @@ def get(self, request, *args, **kwargs):
         json_dict = dict(exists='invalid')
 
         text = request.GET.get('text', '')
+        text = text.strip(' "')
```
nicpottier: Should we just be using a regex to strip anything that isn't a character? Maybe look at how we do matching on RapidPro for this case.. as it deals with UNICODE as well.

ureport/countries/migrations/0002_auto_20150722_1524.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+import re
+
+from django.db import models, migrations
+
+
+def name_stemming(name):
```
nicpottier: function name should be verb.. so ```stem_name``` is better here.

ureport/countries/migrations/0002_auto_20150722_1524.py
```diff
@@ -0,0 +1,31 @@
...
+    words = re.split(r'\W+', name, re.UNICODE)
+    return " ".join([word.lower() for word in words if word])
+
+
+def populated_stemmed_name(apps, schema_editor):
```
nicpottier: Same here.. ```populate_stemmed_names```

nicpottier: Or better yet.. ```stem_country_aliases```

ureport/countries/models.py
```diff
@@ -10,7 +11,14 @@ class CountryAlias(SmartModel):
     name = models.CharField(max_length=128, help_text=_("The name for our alias"))
 
     @classmethod
+    def name_stemming(cls, name):
```
nicpottier: same naming as above.. ```stem_name```

nicpottier: Also.. I have my nomenclature a bit goofed up.. this is more tokenizing than stemming.

so this should be ```tokenize_name``` or maybe ```normalize_name``` even.. honestly ```normalize_name``` is probably most accurate

ureport/public/tests.py
```diff
@@ -1280,3 +1280,21 @@ def test_countries(self):
...
+        response = self.client.get(countries_url + '?text="Etats+Unies"')
+        self.assertEquals(response.status_code, 200)
+        response_json = json.loads(response.content)
+        self.assertTrue('exists' in response_json)
+        self.assertTrue('country_code' in response_json)
+        self.assertEquals(response_json['exists'], "valid")
+        self.assertEquals(response_json['country_code'], "US")
+
+        # country text has quotes an spaces
+        response = self.client.get(countries_url + '?text="    Etats+Unies  "')
```
nicpottier: How about a unicode test?

nicpottier:
Looks good.

PR #55, opened 2015-07-24
Fix login link

norkans7:


templates/frame.haml
```diff
@@ -418,7 +418,7 @@
             .row
               .col-sm-12
                 .login-link
-                  %a.primary-color{href:"{{ org.build_host_link }}/users/login/"}
+                  %a.primary-color{href:"{% org_host_link %}/users/login/"}
```
nicpottier: This is fine for now, but I think the right solution to this would be a template tag where you pass in whether you want the link to be a secure one or not.

norkans7: Ok

PR #58, opened 2015-07-31
make second level for map optional

norkans7:


norkans7:
Disable the click zoom if district label is not set @nicpottier please review

nicpottier:
Looks good, go for it!

PR #65, opened 2015-08-10
Cache refresh

norkans7:


ureport/settings_common.py
```diff
@@ -320,7 +321,7 @@
           'list'),  # can view a list of the objects
 
     'dashblocks.dashblock': ('html', ),
-    'orgs.org': ('choose', 'edit', 'home', 'manage_accounts', 'create_login', 'join'),
+    'org_ext.org': ('choose', 'edit', 'home', 'manage_accounts', 'create_login', 'join', 'refresh_cache'),
```
nicpottier: I don't think this is doing what you think it is doing.

Permissions are per object I believe, and since the root object is Org, there's really nothing to actually set here. I think things are working for you right now because you are using the superuser, but these permissions aren't doing anything at all.

ureport/settings_common.py
```diff
@@ -337,10 +338,10 @@
...
         'news.newsitem.*',
         'news.video.*',
-        'orgs.org_edit',
-        'orgs.org_home',
-        'orgs.org_manage_accounts',
-        'orgs.org_profile',
+        'org_ext.org_edit',
+        'org_ext.org_home',
+        'org_ext.org_manage_accounts',
+        'org_ext.org_profile',
```
nicpottier: Same here, these aren't real permissions I don't think, permissions are on Models, and there is no Org model in org_ext.

ureport/org_ext/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
nicpottier: Suspiciously empty! :)

ureport/org_ext/views.py
```diff
@@ -0,0 +1,65 @@
...
+        def get_success_url(self):
+            return reverse('org_ext.org_home')
+
+    class Home(OrgCRUDL.Home):
+        pass
+
+    class Edit(OrgCRUDL.Edit):
+        pass
+
+    class ManageAccounts(OrgCRUDL.ManageAccounts):
```
nicpottier: Why do these even exist?

ureport/settings_common.py
norkans7: ahh, I see

norkans7:
So I can have just the action of refresh_cache and its view class and still use the dash views for the remaining actions?

nicpottier:
Right.. and maybe that just belongs in a /cache or /admin app or some such.
You can still have the permission live on org and use permissions for it
though.

-Nic

norkans7:
I rework this

PR #66, opened 2015-08-11
add admis app to allow superuser to trigger tasks to rebuild the cache

norkans7:


ureport/admins/views.py
```diff
@@ -0,0 +1,22 @@
...
+class AdminCRUDL(SmartCRUDL):
+    actions = ('refresh_cache',)
+    model = Org
+
+    class RefreshCache(SmartUpdateView):
+        fields = ('id',)
+        success_message = None
+        success_url = '@orgs.org_home'
+
+        def pre_process(self, request, *args, **kwargs):
```
nicpottier: This shouldn't be in pre_process, everything in here is done before permissions are checked etc.. instead it should be in post_save() or the like.

ureport/admins/tasks.py
```diff
@@ -0,0 +1,160 @@
+import time
```
nicpottier: Why did you move all these tasks over from /polls to /admins? 

Seems like the polls app was the more logical place for them.

norkans7: wanted these to be used both by the beat scheduler and the refresh cache view

I can replace them back

ureport/admins/views.py
norkans7: ok

ureport/admins/tasks.py
nicpottier: Ya, to me it feels like tasks that update polls do belong in the /polls app, while the admins app can certainly fire off one of those tasks, it isn't really in charge of polls, the polls app is.

norkans7:
Please review, I think this is ready

ureport/admins/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    main_polls = 2
+    brick_polls = 3
+    other_polls = 4
+    flows = 5
+    reporters = 6
+    graphs_data = 7
+
+
+def refresh_caches(org, caches):
+    if not org:
```
nicpottier: Why would org ever be None?

nicpottier:
Looks good.

ureport/admins/__init__.py
norkans7: right, I remove this, this is unexpected

PR #67, opened 2015-08-12
fix interactivity for maps on global org

norkans7:


nicpottier:
Sure.. maybe it is worth having a checklist for test cases since seems like there are a lot regressions on this stuff? IE, a wiki page that has a set of steps to try before pushing?

norkans7:
Yep, I'll add the checklist

PR #70, opened 2015-08-18
add micro api for extracting flow info from poll url

sypl:
A tiny API for getting `flow_uuid` and title out of the poll ID, which is in the url. We need this for our project: https://github.com/rapidpro/ureport-web-participation.

### DIFF
--- a/ureport/api/views.py
+++ b/ureport/api/views.py
@@ -1,0 +1,19 @@
+from ureport.polls.models import Poll
+from django.http import JsonResponse
+
+
+def get_flow_info(request, poll_id):
+    flow_exists = Poll.objects.filter(id=poll_id).exists()
+    if flow_exists:
+        flow = Poll.objects.get(id=poll_id)
+        return JsonResponse({
+            'poll_id': poll_id,
+            'flow_uuid': flow.flow_uuid,
+            'title': flow.title
+        })
+    else:
+        return JsonResponse({
+            'poll_id': 0,
+            'flow_uuid': 0,
+            'title': None
+        })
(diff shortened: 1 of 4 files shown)


ureport/api/views.py
```diff
@@ -0,0 +1,19 @@
+from ureport.polls.models import Poll
+from django.http import JsonResponse
+
+
+def get_flow_info(request, poll_id):
+    flow_exists = Poll.objects.filter(id=poll_id).exists()
+    if flow_exists:
+        flow = Poll.objects.get(id=poll_id)
```
nicpottier: Probably makes more sense to do:

```python
flow = Poll.objects.filter(id=poll_id).first()
if flow:
```

ureport/api/views.py
```diff
@@ -0,0 +1,19 @@
...
+        return JsonResponse({
+            'poll_id': poll_id,
+            'flow_uuid': flow.flow_uuid,
+            'title': flow.title
+        })
+    else:
+        return JsonResponse({
```
nicpottier: Returning a 404 is the REST way to handle this.

xkmato:
There are a couple of other things that we are planning to expose on ureport i.e boundary assets, programme news and articles, etc and I the plan is to build an entire API to expose most of that.

sypl:
Do you have a timeline for that? Our project is has about five weeks left, and we need at least this much in order for our online polling to work.

Looking ahead to our other stories though there are other things we need exposed, so if there was a framework for an API then that would be great.

nicpottier:
I definitely prefer having a full API using @xkmato's approach. I don't think getting that in this week would be a big deal.

sypl:
I took a look at that and it looks better than mine. Looking forward to
using it.

norkans7:
The API has been implemented in #73

PR #73, opened 2015-08-19
Exposing public Ureport entities

xkmato:
This is a read only API. Exposing all the public entities; orgs, poll, news items, videos, assets and stories. There is also #70 that is trying to expose flow uuids.


ureport/api/views.py
```diff
@@ -0,0 +1,97 @@
...
+__author__ = 'kenneth'
+
+
+class TokenCRUDL(SmartCRUDL):
+    model = Token
+    actions = ('list', )
+
+    class List(SmartListView):
+        search_fields = ('user__username__icontains','user__first_name__icontains', 'user__last_name__icontains')
+        fields = ('user', 'key')
```
nicpottier: What purpose does this have? And isn't this a leak of all user data?

ureport/api/urls.py
```diff
@@ -0,0 +1,32 @@
...
+                       url(r'^orgs/$', OrgList.as_view()),
+                       url(r'^orgs/(?P<pk>[\w]+)/$', OrgDetails.as_view()),
+
+                       url(r'^polls/$', PollList.as_view()),
+                       url(r'^polls/org/(?P<org>[\w]+)/$', PollList.as_view()),
+                       url(r'^polls/(?P<pk>[\w]+)/$', PollDetails.as_view()),
+
+                       url(r'^news-items/$', NewsItemList.as_view()),
+                       url(r'^news-items/org/(?P<org>[\w]+)/$', NewsItemList.as_view()),
+                       url(r'^news-items/(?P<pk>[\w]+)/$', NewsItemDetails.as_view()),
```
nicpottier: no dashes, just do ```newsitems```

ureport/api/urls.py
```diff
@@ -0,0 +1,32 @@
...
+
+urlpatterns = patterns('',
+                       url(r'^orgs/$', OrgList.as_view()),
+                       url(r'^orgs/(?P<pk>[\w]+)/$', OrgDetails.as_view()),
+
+                       url(r'^polls/$', PollList.as_view()),
+                       url(r'^polls/org/(?P<org>[\w]+)/$', PollList.as_view()),
+                       url(r'^polls/(?P<pk>[\w]+)/$', PollDetails.as_view()),
+
+                       url(r'^news-items/$', NewsItemList.as_view()),
```
nicpottier: I would rather remove all the top level polls, news, videos, assets and stories endpoints and always force an Org id. I don't see any need for a client to get all at once.

ureport/api/tests.py
```diff
@@ -0,0 +1,158 @@
...
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['count'], Org.objects.count())
+
+    def test_polls_list(self):
+        url = '/api/v1/polls/'
+        self.client.login(username=self.superuser.username, password='super')
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data['count'], Poll.objects.count())
```
nicpottier: Good job on having tests, but these should be testing the actual serialization as well.

nicpottier:
I'm  good with this if my comments are addressed.

ureport/api/views.py
xkmato: My plan here was to create a place where we can show user tokens. But I think the alternative could be putting it in dash's /users/user/profile/<id>/ 
So I am taking it out while I figure it out

xkmato: Also this would matter if we want people/apps to authenticate to access the API, which I am also not sure is very relevant

ureport/api/serializers.py
```diff
@@ -0,0 +1,97 @@
...
+            return generate_absolute_url_from_file(self.context['request'], image)
+        return None
+
+
+class OrgReadSerializer(serializers.ModelSerializer):
+    logo_url = SerializerMethodField()
+
+    class Meta:
+        model = Org
+        exclude = ('administrators', 'viewers', 'editors', 'api_token', 'config', 'logo','created_on', 'modified_on',
```
norkans7: It is preferred to explicitly lis the fields instead of exclude

xkmato: More predictable, makes sense

ewheeler:
hey @nicpottier @norkans7 
i believe @xkmato's recent commits address all comment. 

pls take another look and let me know when we might be able to see this deployed (our partner is blocked and is anxiously awaiting this API)

ureport/api/serializers.py
```diff
@@ -0,0 +1,97 @@
...
+def generate_absolute_url_from_file(request, file):
+    return request.build_absolute_uri(file.url)
+
+
+class CategoryReadSerializer(serializers.ModelSerializer):
+    image_url = SerializerMethodField()
+
+    class Meta:
+        model = Category
+        fields = ('image_url', 'is_active', 'name',)
```
norkans7: I think this API is to read the data that are shown on the public site, I think is_active fields is probably not needed and we should make sure the object exposed are filtered by is_active=True.  Unless it is useful to see the inactive objects

ureport/api/serializers.py
```diff
@@ -0,0 +1,97 @@
...
+        return None
+
+
+class StoryReadSerializer(serializers.ModelSerializer):
+    category = CategoryReadSerializer()
+    images = SerializerMethodField()
+
+    class Meta:
+        model = Story
+        fields = ('id', 'is_active', 'title', 'featured', 'summary', 'video_id', 'tags', 'org', 'images', 'category')
```
norkans7: On Story we have a new field for audio_link, I think you can include that as well

again I guess is_active is not needed

ureport/api/views.py
```diff
@@ -0,0 +1,86 @@
...
+from ureport.assets.models import Image
+from ureport.news.models import NewsItem, Video
+from ureport.polls.models import Poll
+
+__author__ = 'kenneth'
+
+
+class OrgList(ListAPIView):
+    serializer_class = OrgReadSerializer
+    queryset = Org.objects.all()
```
norkans7: I think the user should calling the API should list the org he is part of

ureport/api/views.py
```diff
@@ -0,0 +1,86 @@
...
+    def get_queryset(self):
+        q = super(PollList, self).get_queryset()
+        if self.request.query_params.get('flow_uuid', None):
+            q = q.filter(flow_uuid=self.request.query_params.get('flow_uuid'))
+        return q
+
+
+class PollDetails(RetrieveAPIView):
+    serializer_class = PollReadSerializer
+    queryset = Poll.objects.all()
```
norkans7: filter by user org and is_active=True, I do not think we need to allow users to be able to request a poll on an org they are not part of, same for all objects below

norkans7:
Also have a way to show the user their API token somewhere

xkmato:
I just took out Authentication, so all that is going to be public. I am wondering why we want restrict what orgs a user can see?

xkmato:
Hey @norkans7 I have filtered everything by is_active and taken out is_active from the item fields. 

@ewheeler @nicpottier I am wondering what the stand on the whole authentication thing is. Personally I don't see the point of authentication, since all of this stuff is public already.

nicpottier:
Functionally this looks good, agreed that we don't need authentication and that returning all the orgs is fine.

Still missing docs though. @xkmato I believe Django-Rest-Frameowork will generate nice docs off of pydocs formatting on the endpoints. See "self describing APIs" here:
  http://www.django-rest-framework.org/topics/documenting-your-api/

xkmato:
I think this guy is now good to go. @norkans7 @ewheeler @nicpottier whenever you get time.

nicpottier:
Nice, let's do this.

@norkans7 Can you push this today?

norkans7:
sure

PR #75, opened 2015-08-26
Translations

norkans7:


ureport/urls.py
```diff
@@ -19,6 +24,7 @@
     url(r'^manage/', include('ureport.countries.urls')),
     url(r'^manage/', include('ureport.assets.urls')),
     url(r'^users/', include('dash.users.urls')),
+    url(r'^jsi18n/$', 'django.views.i18n.javascript_catalog', js_info_dict)
```
nicpottier: Have you tested this? Doesn't this need to be included in a template now?

norkans7: sadly NO

nicpottier:
Looks like dash and smartmin aren't included.. do you have those symlinked?

norkans7:
Okay I had only dash symlink, I can add smartmin as well

nicpottier:
Ya, what I look for in these is big patches of things being removed.. in this case all the smartmin translations.. which would have been sad. :(

norkans7:
I was a bit confused before too, now this looks sane

nicpottier:
Nice, looks good.

Ok next step is to upload the English .po file to Transifex. We can then get Maria to update the Spanish translations that are missing. 

Once she is done with that, then we download the .po files from Transifex, then do a ```fab compilemessages``` and push.

PR #77, opened 2015-08-28
add parameter to poll listings to display only latest

sypl:
Would it be possible to get this parameter added to the api? I need to be able to know what the latest poll is for a particular org.

### DIFF
--- a/ureport/api/views.py
+++ b/ureport/api/views.py
@@ -101,6 +101,7 @@ class PollList(BaseListAPIView):
     * **title** - the TITLE of the poll (string)
     * **org** - the ID of the org that owns this poll (int)
     * **category** - the CATEGORIES of of this poll (dictionary)
+    * **latest** - the latest poll, if set to 'true' (dictionary)
 
     Example:
 
@@ -133,6 +134,13 @@ class PollList(BaseListAPIView):
         q = super(PollList, self).get_queryset()
         if self.request.query_params.get('flow_uuid', None):
             q = q.filter(flow_uuid=self.request.query_params.get('flow_uuid'))
+        if self.request.query_params.get('latest', None):
+            if self.request.query_params.get('latest') == 'true':
+                org = self.kwargs.get('org')
+                main_poll = Poll.get_main_poll(org)
+                latest_poll_id = main_poll.id
+                q = q.filter(id=latest_poll_id)
+
         return q
 
 
@@ -461,5 +469,3 @@ class StoryDetails(RetrieveAPIView):
     """
     serializer_class = StoryReadSerializer
     queryset = Story.objects.filter(is_active=True)
-
-
--- a/ureport/api/views.py
+++ b/ureport/api/views.py
@@ -138,8 +138,9 @@ class PollList(BaseListAPIView):
             if self.request.query_params.get('latest') == 'true':
                 org = self.kwargs.get('org')
                 main_poll = Poll.get_main_poll(org)
-                latest_poll_id = main_poll.id
-                q = q.filter(id=latest_poll_id)
+                if main_poll:
+                    latest_poll_id = main_poll.id
+                    q = q.filter(id=latest_poll_id)
 
         return q
 
--- a/ureport/api/views.py
+++ b/ureport/api/views.py
@@ -101,7 +101,6 @@ class PollList(BaseListAPIView):
     * **title** - the TITLE of the poll (string)
     * **org** - the ID of the org that owns this poll (int)
     * **category** - the CATEGORIES of of this poll (dictionary)
-    * **latest** - the latest poll, if set to 'true' (dictionary)
 
     Example:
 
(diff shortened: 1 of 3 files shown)


nicpottier:
I'd rather see this be 'featured' rather than latest, because I think that's what you are really after.

Also, unit test!

sypl:
Sure thing. Will write it over the weekend.

xkmato:
Also if it is always going to be just the one poll why not give it as an id in poll detail `/api/v1/polls/featured/` that would return `main_poll` instead of always returning a list of one item

sypl:
Fair point. I'll see if I can make that happen.

sypl:
I've implemented this as a list view. I tried to list as a detail view under `/api/v1/polls/featured`, but there's no way to pass in the `org_id` that way.

sypl:
Here's my attempt at featured poll as a detail view: https://github.com/sypl/ureport/commit/b0a880eb5754bf7e4c97933286a20c84cb9aeeb9

Getting a 404 for that url. Not sure what the queryset wants returned. A list of Polls I think, but that's not working. If anyone has any ideas I'd love to hear them. Or, if the list implementation is ok, then we can call it a day.

nicpottier:
Whoa, competing pull requests!

So I actually think a list on /featured is probably the right thing, as you have it Lui. The one tweak I would make is to actually put all the featured polls, in reverse chronological order there, for consistency with naming.

Looking at the implementation of ```get_main_poll``` it looks like it backs down to the most recent poll if there are no featured polls:
 https://github.com/rapidpro/ureport/blob/9f8cdf765a2411f82849189f4396c4c9e3dfe44a/ureport/polls/models.py#L90

So I would do something like:
1) Get ```Poll.objects.filter(featured=True).order_by('-created_on')```, if not an empty list, return that
2) If 1) returns an empty list, then get_main_poll() and return that poll (as a list of 1)

I could also be convinced that 2) should actually return an empty list (because that poll isn't featured) and instead leave it to the callers to implement that logic of getting the most recent poll with questions, but I'm fine with either.

@xkmato This PR looks fine to me with the above changes.

ureport/api/views.py
```diff
@@ -101,6 +101,7 @@ class PollList(BaseListAPIView):
     * **title** - the TITLE of the poll (string)
     * **org** - the ID of the org that owns this poll (int)
     * **category** - the CATEGORIES of of this poll (dictionary)
+    * **latest** - the latest poll, if set to 'true' (dictionary)
```
nicpottier: Left over from previous implementation.

ureport/api/views.py
```diff
@@ -163,6 +164,49 @@ class PollDetails(RetrieveAPIView):
...
+    """
+    This endpoint allows you to list the featured poll.
+
+    ## Listing Featured Poll
+
+    Example:
+
+        GET /api/v1/polls/org/1/featured/
+
+    Response is a list with the single featured poll, or an empty results list if none is featured:
```
nicpottier: Response is a list of the featured polls, most recent first. An empty list is returned if there are no polls with questions.

xkmato:
I agree, @nicpottier having a list of all featured polls makes a lot of sense, it would cover more scenarios that just getting the main poll. I had not clear looked through the logic for returning the main poll, I thought it was more special than other featured polls.

sypl:
My motivation for this endpoint is that I need to link to the latest poll,
however we decide to identify that. As Poll has the get_main_poll method, I
thought it convenient to re-use that. Listing all featured polls doesn't
really solve my problem, as there's no mechanism for identifying which is
'the one'. Actually, as Nic mentioned, the get_main_poll method falls back
to most recent poll if none are featured, but I think that's ok as the dev
can just check that poll to see if it has the is_featured attribute if
they're not interested in the fallback.

I personally would prefer Kenneth's new pull request, sending just the
details of the single poll identified as 'main'. I could live with a list
of featured polls for all organisations, as long as each organisation only
has one poll, and I can filter out my desired organisation via a parameter,
so something like this: /api/v1/polls/featured/?org=1

nicpottier:
You don't mean latest though, you mean the poll shown on the front page right?

From an API point of view I much prefer sending back a list of the polls that have been featured and having you select the top one. It is generally better to provide functionality that will be useful to all, not just your one use case, and in this case it is just as easy for you to just grab the first result off of /featured

sypl:
The plan that we've discussed with Unicef is that we add a 'Take Latest Poll' button on the homepage. Taking the latest featured poll for an org gives me that, so I can live with it. It seems a bit of a duplication of the `get_main_poll` logic, but I do see the benefits of the more general approach.

So just to be sure we're talking about the same thing, we're talking about this endpoint:
`/api/v1/polls/featured/` returning a list of all featured polls for all organisations, listed in reverse chronological order, correct?

nicpottier:
Am I reading your PR wrong? I thought you added a: ```/api/v1/polls/[org_id]/featured``` endpoint?

Having it broken out by org definitely makes sense to me.

sypl:
Glad I clarified then ;) Yeah, sure, by organisation is fine, but the
original endpoint was `polls/org/(?P<org>[\w]+)/featured/`. That's ok right?

nicpottier:
Yep yep.

sypl:
Alright, I think that'll do it. Lemme know if anything is awry.

nicpottier:
Looks good to me. @norkans7 can you review, merge and ship?

norkans7:
Looks good

sypl:
Thanks guys

norkans7:
Already pushed this

PR #78, opened 2015-08-31
Api featured poll

xkmato:
This is for displaying the featured poll of an organisation. Request for #77


xkmato:
@sypl I am thinking this is what you had in mind @norkans7 @nicpottier Please review

xkmato:
#77 Has this covered

PR #79, opened 2015-09-02
Adding questions to polls endpoint

xkmato:
Although most of this derivable from the flow_uuid, the team working on the app think that it is more efficient and convenient for them to get the cached `get_total_summary_data` from Ureport. That is the main reason I have added questions to this endpoint. Please check it out and let me know what you think. 

* Overrode the drf template for the browsable API. 
* Added time stamps on Polls, News, Stories, Assets endpoints
* Added questions to polls

### DIFF
--- a/ureport/api/serializers.py
+++ b/ureport/api/serializers.py
@@ -5,7 +5,7 @@ from rest_framework import serializers
 from rest_framework.fields import SerializerMethodField
 from ureport.assets.models import Image
 from ureport.news.models import NewsItem, Video
-from ureport.polls.models import Poll
+from ureport.polls.models import Poll, PollQuestion
 
 __author__ = 'kenneth'
 
@@ -19,7 +19,7 @@ class CategoryReadSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Category
-        fields = ('image_url', 'name',)
+        fields = ('image_url', 'name', )
 
     def get_image_url(self, obj):
         image = None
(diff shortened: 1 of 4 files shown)


ureport/api/serializers.py
```diff
@@ -51,19 +51,32 @@ class StoryReadSerializer(serializers.ModelSerializer):
...
         return [generate_absolute_url_from_file(self.context['request'], image.image)
                 for image in obj.get_featured_images()]
 
 
+class PollQuestionReadSerializer(serializers.ModelSerializer):
+    total_summary_data = SerializerMethodField()
+
+    class Meta:
+        model = PollQuestion
+        fields = ('title', 'ruleset_uuid', 'is_open_ended', 'total_summary_data', )
```
norkans7: the way we get the data from RapidPro is changing soon may be for now do not include the 'total_summary_data'

ureport/api/tests.py
```diff
@@ -158,10 +158,13 @@ def test_single_poll(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         poll = self.reg_poll
         category = response.data.pop('category')
+        print response.data
```
norkans7: print

ureport/api/tests.py
```diff
@@ -188,7 +191,8 @@ def test_single_news_item(self):
                                                  title=news.title,
                                                  description=news.description,
                                                  link=news.link,
-                                                 org=news.org_id))
+                                                 org=news.org_id,
+                                                 created_on=news.created_on.strftime('%Y-%m-%dT%H:%M:%S.%fZ')))
```
norkans7: for consistency here you can use https://github.com/nyaruka/rapidpro/blob/master/temba/utils/__init__.py#L110
to just consider the first 3 digits in microseconds
I have added it in utils on a branch I am working on in utils

xkmato: Great, let me wait for the branch to ship.

ureport/api/serializers.py
xkmato: I think, I will just wait for the new way. Because the whole point of this PR was to expose total_summary_data so the Android client does not have to calculate it

nicpottier:
Thought I had replied to those guys that this was undergoing changes at the minute. We are redoing how all of these stats are calculated on U-Report, so it doesn't make sense to expose these just yet. That change is a few weeks out though.

norkans7:
Sorry @xkmato I had forgotten about this PR, and I was implementing https://github.com/rapidpro/ureport/pull/140/commits/a9120c19eebf97fec97e9672af05f4a93355816c

Since this PR has more other nice things to have, what about you merge master and update it so we can merge this instead of that commit?

ureport/api/serializers.py
norkans7: I think here the fields can be title, ruleset_uuid, id and results and use get_results method, to me total_summary_data is a bit confusing on the API,

norkans7:
pulled in #140 so closing this

PR #80, opened 2015-09-09
use written_by field on story page

norkans7:
Depends https://github.com/rapidpro/dash/pull/51


norkans7:
@nicpottier plz review these 2 PR

nicpottier:
Seems could be cleaner to have a ```get_written_by``` or something on Story that does the switch, then you can easily test this and have no if blocks in your template.

norkans7:
right

PR #81, opened 2015-09-18
Reporters trigger 1

norkans7:


ureport/contacts/migrations/0004_auto_20150918_0915.py
```diff
@@ -0,0 +1,116 @@
...
+TRIGGER_SQL = """
+-----------------------------------------------------------------------------
+-- Increment or decrement the reporters count on counter of given type
+-----------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION
+  ureport_insert_reporters_counter(_org_id INT, _type VARCHAR, _count INT)
+RETURNS VOID AS $$
+BEGIN
+  INSERT INTO contacts_reporterscounter("org_id", "type", "count") VALUES(_org_id, _type, _count);
+  PERFORM ureport_may_be_squash_reporterscounters(_org_id, _type);
```
nicpottier: ```maybe``` is one word.

norkans7: ok

ureport/contacts/migrations/0004_auto_20150918_0915.py
```diff
@@ -0,0 +1,116 @@
...
+      RETURNING "count")
+      INSERT INTO contacts_reporterscounter("org_id", "type", "count")
+      VALUES (_org_id, _type ,GREATEST(0, (SELECT SUM("count") FROM deleted)));
+  END IF;
+END;
+$$ LANGUAGE plpgsql;
+-----------------------------------------------------------------------------
+-- Increment counters for a contact
+-----------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION ureport_increment_counter_for_contact(_contact contacts_contact, _add BOOLEAN)
```
nicpottier: This approach seems like a lot more updates than strictly necessary. You are decrementing and incrementing all possible field values for any contact change?

Couldn't you do this based on whether the field changed in an update/insert?

nicpottier:
This looks really good though I think you should try tweaking the triggers so we don't have to do full decrement and increment of counts on every contact field, instead just doing those that have changed.

ureport/contacts/migrations/0004_auto_20150918_0915.py
norkans7: Ok, I'll add another method to use on UPDATE that considers the values on OLD and NEW

nicpottier:
Go for it.

PR #84, opened 2015-09-22
update response rate to consider the contacts that responded to at le…

norkans7:
…ast the first question


nicpottier:
Buys us some time, so sure. :)

norkans7:
yep

PR #90, opened 2015-10-29
Rtl ui

norkans7:


norkans7:
Please Review

templates/public/index.haml
```diff
@@ -675,7 +711,11 @@
       $(function(){
         $("#age-chart").each(function(){
 
-           var data = {{ age_stats|safe }}
+           var data = {{ age_stats|safe }};
+
+           {% if is_rtl_org %}
+           data.reverse();
```
nicpottier: clever

templates/public/index.haml
```diff
@@ -325,7 +349,10 @@
                   {{ story.title }}
                 %a.brick-link.primary-bg-color{ href:"{% url 'public.story_read' story.pk %}" }
                   - trans "READ FULL STORY"
-                  &#8594;
+                  -if is_rtl_org
+                    &#8592;
+                  -else
+                    &#8594;
```
nicpottier: Instead of so many if's.. might it make sense to have a template tag that resolves to the right thing here based on the org?

nicpottier:
Nice.. I would just figure out how not to repeat that if block and then looks good!

PR #91, opened 2015-11-11
order flows alphabetically to facilitate selection and include creati…

norkans7:
…on date in the choice display

### DIFF
--- a/ureport/polls/tests.py
+++ b/ureport/polls/tests.py
@@ -468,7 +468,7 @@ class PollTest(DashTest):
         with patch('dash.orgs.models.Org.get_flows') as mock:
             flows_cached = dict()
             flows_cached['uuid-25'] = dict(runs=300, completed_runs=120, name='Flow 1', uuid='uuid-25', participants=300,
-                                           labels="", archived=False, created_on="2015-04-08T12:48:44.320Z",
+                                           labels="", archived=False, created_on="2015-04-08",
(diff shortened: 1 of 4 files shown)


ureport/polls/tests.py
```diff
@@ -468,7 +468,7 @@ def test_create_poll(self):
         with patch('dash.orgs.models.Org.get_flows') as mock:
```
rowanseymour: Am confused by this. Where is `get_flows` defined? I don't see it in https://github.com/rapidpro/dash/blob/master/dash/orgs/models.py

norkans7: those methods are really specific to ureport so there are not is dash but https://github.com/rapidpro/ureport/blob/order_flows/ureport/utils/__init__.py#L331

and they lookup the data from the cache which is updated by a task by calling 
https://github.com/rapidpro/ureport/blob/order_flows/ureport/utils/__init__.py#L292

ureport/utils/__init__.py
```diff
@@ -306,6 +306,7 @@ def fetch_flows(org):
             if flow.rulesets:
                 flow_json = dict()
                 flow_json['uuid'] = flow.uuid
+                flow_json['created_on'] = flow.created_on.strftime('%Y-%m-%d')
```
rowanseymour: If you think you might ever need the full `created_on` you could store it in the cache as full iso string and reduce it to just the date when rendering the PollForm

rowanseymour:
Looks fine

PR #92, opened 2015-11-11
Pre poll results triggers to store other data than counts not relying on cache results

norkans7:


ureport/polls/migrations/0024_populate-response-categories.py
```diff
@@ -0,0 +1,95 @@
...
+                poll.save()
+
+                deactivated_ids.append(poll.pk)
+                deactivated += 1
+                print "Hidden poll - %s - %d - %s" % (org.name, poll.pk, poll.title)
+
+            except Exception as e:
+                raise e
+
+        print "Finished populating %d polls in %ss" % (successes, time.time() - start)
```
norkans7: this migration takes around 5 min and deactivate 1 poll

ureport/polls/tests.py
```diff
@@ -667,79 +658,62 @@ def test_questions_poll(self):
                                                      created_by=self.admin,
                                                      modified_by=self.admin)
 
-        with patch('ureport.polls.models.Poll.get_flow') as mock:
```
norkans7: We do not need to get the poll flow the question are created in post save of create or update 
so here it is just the decision to show them or leave them hidden

PR #94, opened 2015-11-17
Update context processors to populate background images

norkans7:


norkans7:
Please review

nicpottier:
This feels like a different thing to me than the set_org context processor. My preference would probably to keep the dash context processor but then add the background image stuff as a separate processor.

Issue #98, opened 2015-12-14
ImportError: cannot import name get_redis_connection


EswarSakamuri:
Traceback (most recent call last):
  File "manage.py", line 8, in <module>
    execute_from_command_line(sys.argv)
  File "/home/administrator/workspace/ureportenv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/home/administrator/workspace/ureportenv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 312, in execute
    django.setup()
  File "/home/administrator/workspace/ureportenv/local/lib/python2.7/site-packages/django/__init__.py", line 18, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/home/administrator/workspace/ureportenv/local/lib/python2.7/site-packages/django/apps/registry.py", line 85, in populate
    app_config = AppConfig.create(entry)
  File "/home/administrator/workspace/ureportenv/local/lib/python2.7/site-packages/django/apps/config.py", line 119, in create
    import_module(entry)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/home/administrator/workspace/ureport/ureport/admins/__init__.py", line 1, in <module>
    from dash.orgs.models import Org
  File "/home/administrator/workspace/ureportenv/src/dash/dash/orgs/models.py", line 18, in <module>
    from dash.api import API
  File "/home/administrator/workspace/ureportenv/src/dash/dash/api/__init__.py", line 6, in <module>
    from redis_cache import get_redis_connection
ImportError: cannot import name get_redis_connection

xkmato:
My guess is that you're using the wrong or outdated python redis lib. Make sure you env's pip freeze output is the same as the contents of pip-freeze.txt

EswarSakamuri:
With pip-freeze.txt its working fine I got previous error when installed with pip-requires.txt

one more thing when creating a symbolic link using 
ln -s ureport/settings.py.postgres ureport/settings.py

python manage.py syncdb giving me this error 
ImportError: No module named settings(Did I make any mistake in running the sync command?)

at present I created a fresh copy of settings.py from settings_common.py

xkmato:
Just make a copy instead of link do `cp ureport/settings.py.postgres ureport/settings.py` instead

norkans7:
Symlink is the right thing instead of copy, even if either will work

[...]

PR #99, opened 2015-12-14
Adjusting readme to create more platform consistent instructions

xkmato:
Making simple changes in the README so that issues like #98 don't appear as often as they do.
- Change Install dependencies from `pip install -r pip-requires.txt` to `pip install pip-freeze.txt`
- Change creating settings file from creating sym link to creating copy of file. symbolics links don't seem to have consistent behaviour on different filesystems or platforms

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,12 @@ Install dependencies
 ```
 % virtualenv env
 % source env/bin/activate
-% pip install -r pip-requires.txt
+% pip install -r pip-freeze.txt
 ```
 
 Link up a settings file (you'll need to create the postgres db first, username: 'ureport' password: 'nyaruka')
 ```
-% ln -s ureport/settings.py.postgres ureport/settings.py
+% cp ureport/settings.py.postgres ureport/settings.py
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,12 +17,12 @@ Install dependencies
 ```
 % virtualenv env
 % source env/bin/activate
-% pip install -r pip-requires.txt
+% pip install -r pip-freeze.txt
 ```
 
 Link up a settings file (you'll need to create the postgres db first, username: 'ureport' password: 'nyaruka')
 ```
-% ln -s ureport/settings.py.postgres ureport/settings.py
```
norkans7: Here the symlink is the right thing

xkmato: Well links have different behaviours on different OS. On many platforms, links are not recognised as python modules. Also I think copying the file would give one the freedom to adjust the settings without affecting the repo since `settings.py` is ignored, otherwise for the link, you always have to adjust the linked-to file which is usually in the repo

norkans7:
I think symlink is the right thing. Closing this

PR #100, opened 2015-12-15
Reduce lock

norkans7:


nicpottier:
Why not keep the count link until Uganda goes live? Do we include the counts of non-active orgs?

norkans7:
once the backfill is done, we push this and make it active,

nicpottier:
Ok that works

norkans7:
replaced by #101

PR #104, opened 2016-01-05
Expose Org stats in api

xkmato:
There is need to expose the stats that are shown on /ureporters page in API.

Writing tests for this is a bit challenging since the results for these are not generated but read from cache instead. If anyone has any idea has any ideas on how I can go about testing these, I will appreciate very much.

### DIFF
--- a/ureport/api/tests.py
+++ b/ureport/api/tests.py
@@ -124,10 +124,10 @@ class UreportAPITests(APITestCase):
                                                  subdomain=org.subdomain,
                                                  domain=org.domain,
                                                  timezone=org.timezone))
-        self.assertEquals(gender_stats, org.get_gender_stats)
-        self.assertEquals(age_stats, org.get_age_stats)
-        self.assertEquals(registration_stats, org.get_registration_stats)
-        self.assertEquals(occupation_stats, org.get_occupation_stats)
+        self.assertEquals(gender_stats, org.get_gender_stats())
+        self.assertEquals(age_stats, org.get_age_stats())
+        self.assertEquals(registration_stats, org.get_registration_stats())
+        self.assertEquals(occupation_stats, org.get_occupation_stats())
 
     def test_polls_by_org_list(self):
         url = '/api/v1/polls/org/%d/' % self.uganda.pk
--- a/ureport/api/tests.py
+++ b/ureport/api/tests.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+import json
 from random import randint
 from dash.categories.models import Category
 from dash.orgs.models import Org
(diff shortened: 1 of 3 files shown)


norkans7:
@xkmato I think you can test them the same way we test those method here 
https://github.com/rapidpro/ureport/blob/master/ureport/utils/tests.py#L514

xkmato:
Thanks @norkans7 I have decided to compare the API results to the ORG fields since they're already tested anyway. Take a look

ureport/api/views.py
```diff
@@ -45,6 +45,15 @@ class OrgList(ListAPIView):
                 "subdomain": "test",
                 "domain": "ureport.in",
                 "timezone": "Africa/Kampala"
+                "gender_stats": {
+                "female_count": 0,
+                "male_percentage": "---",
+                "female_percentage": "---",
+                "male_count": 0
+                },
+                "age_stats": "[]",
```
norkans7: I think it is better if these are not dumps of the value so [] instead of "[]"

ureport/api/views.py
```diff
@@ -45,6 +45,15 @@ class OrgList(ListAPIView):
...
+                "registration_stats": "[{\"count\": 0, \"label\": \"07/06/15\"}]",
```
norkans7: same here have a list of dict instead of its json dump

ureport/api/tests.py
```diff
@@ -120,6 +124,10 @@ def test_single_org(self):
                                                  subdomain=org.subdomain,
                                                  domain=org.domain,
                                                  timezone=org.timezone))
+        self.assertEquals(gender_stats, org.get_gender_stats())
```
norkans7: I think is is better to test with values so even if we change these methods we can still try to keep the API endpoints the same to not break other people implementations

ureport/api/views.py
xkmato: Yes nice catch here

ureport/api/views.py
xkmato: :+1:

ureport/api/tests.py
xkmato: Since this exposes that method as is, I don't see how it can be tested independent of the method. I see the only way to test with values would be to duplicate https://github.com/rapidpro/ureport/blob/master/ureport/utils/tests.py#L514 - Do you have any suggestions of how else I can go about it?

norkans7: Yep, I think it is fine to have those again, you create ReportersCounter objects and assert the expected values
```python
self.assertEquals(gender_stats, dict(male=35, female=55))
```
instead of 
```python
self.assertEquals(gender_stats, org.get_gender_stats())
```
that will assure us that the API is kept consistent even if something is changed on the methods, we'll have to update the API part to stay the same as before.

xkmato: :+1:

xkmato:
@norkans7 Check it out when you get time

norkans7:
Looks good, Thanks @xkmato

norkans7:
@xkmato this is live now

xkmato:
Thank you

PR #106, opened 2016-01-08
Synchronization of removed contacts on RapidPRO

johncordeiro:
This update fixes the synchronization of removed contacts on RapidPRO side. The error occurs when someone deletes a contact on RapidPRO or from the main U-Reporters group, in this case the amount of U-Reporters on the RapidPRO keeps different from U-Report dashboard.

### DIFF
--- a/ureport/contacts/models.py
+++ b/ureport/contacts/models.py
@@ -249,20 +249,19 @@ class Contact(models.Model):
 
         return seen_uuids
 
-	@classmethod
-	def sync_contacts_removed(cls, org, temba_contacts):
-		ureport_contacts = cls.objects.all()
+    @classmethod
+    def sync_contacts_removed(cls, org, temba_contacts):
+        ureport_contacts = cls.objects.all()
 
-		contacts_removed = []
+        contacts_removed = []
 
-		for ureport_contact in ureport_contacts:
+        for ureport_contact in ureport_contacts:
             if ureport_contact.uuid not in temba_contacts:
                 ureport_contact.delete()
                 ReportersCounter(org=org, type='total-reporters', count=-1)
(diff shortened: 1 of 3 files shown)


ureport/contacts/tasks.py
```diff
@@ -50,7 +50,8 @@ def fetch_contacts_task(org_id=None, fetch_all=False):
 
                     Boundary.get_boundaries(org)
                     ContactField.get_contact_fields(org)
-                    Contact.fetch_contacts(org, after=after)
+                    contacts = Contact.fetch_contacts(org, after=after)
+                    Contact.sync_contacts_removed(org, contacts)
```
norkans7: I guess this should be only if after is None otherwise we will be deleting all contacts added before date after

ureport/contacts/models.py
```diff
@@ -249,9 +249,25 @@ def fetch_contacts(cls, org, after=None):
...
+    def sync_contacts_removed(cls, org, temba_contacts):
 
-class ReportersCounter(models.Model):
+        contacts_removed = []
+        try:
+            contacts_removed = Contact.objects.all().exclude(uuid__in=temba_contacts)
+
+            for contact in contacts_removed:
+                contact.delete()
+                ReportersCounter(org=org, type='total-reporters', count=-1)
```
norkans7: I think this model creation is not needed and not doing anything here,
we update the counts by postgres triggers https://github.com/rapidpro/ureport/blob/dc10f382cb63aa6c4e0d9e3c550802450f332cd6/ureport/contacts/migrations/0004_auto_20150918_0915.py#L7

norkans7:
This functionality has been included in #125

PR #107, opened 2016-01-14
Fix most active regions

norkans7:
On the home page we should display the 4 top most active regions below the reporters counts


norkans7:
@nicpottier please review

nicpottier:
Yay for more tests!

PR #108, opened 2016-01-15
change age brackets

norkans7:


ureport/utils/__init__.py
```diff
@@ -360,18 +360,35 @@ def get_age_stats(org):
...
     age_counts_interval = dict()
+    age_counts_interval['0-14'] = 0
+    age_counts_interval['15-19'] = 0
+    age_counts_interval['20-24'] = 0
+    age_counts_interval['25-30'] = 0
+    age_counts_interval['31-34'] = 0
+    age_counts_interval['35+'] = 0
+
     total = 0
     for year_key, age_count in age_counts.iteritems():
```
nicpottier: Might be clearer to name this year_counts instead of age_counts, as I believe it is representing how many contacts have that birth year right?

nicpottier:
Looks good apart from slightly confusing variable naming.

PR #110, opened 2016-01-17
refetch old sites if not cached values

norkans7:


norkans7:
to assure global counter do not drop

nicpottier:
Yep.

PR #82, opened 2015-09-21
Reporters trigger 2

norkans7:
Should be merged after #81 is pushed

PR #111, opened 2016-01-19
Updated system to show extra admin levels

musamusa:


xkmato:
@norkans7 Please take a look whenever you get time. Cheers

static/less/ureport2.less
```diff
@@ -1809,15 +1811,29 @@ ul.poll-filters, .featured-stories-filters, ul.stories-filters {
...
       top:6px;
     }
+  }
 
+  .scale-map-vline-2 {
+    width:6px;
+    border-left: 6px solid #FFFFFF;
+    height: 14px;
+    position: absolute;
+    left: 280px;
```
norkans7: I found it is better to use right on the last level since with localization the word can be long; may be use the center styles for the middle one instead of hard left values

static/less/ureport2.less
```diff
@@ -1740,16 +1740,18 @@ ul.poll-filters, .featured-stories-filters, ul.stories-filters {
     }
```
norkans7: can you add screenshot for the styles change

ureport/public/views.py
```diff
@@ -208,10 +208,9 @@ def render_to_response(self, context, **kwargs):
         if org.get_config('is_global'):
             location_boundaries = org.boundaries.filter(level=0)
         else:
-            state_id = self.kwargs.get('osm_id', None)
-
-            if state_id:
-                location_boundaries = org.boundaries.filter(level=2, parent__osm_id=state_id)
+            parent_id = self.kwargs.get('osm_id', None)
+            if parent_id:
+                location_boundaries = org.boundaries.filter(parent__osm_id=parent_id)
```
norkans7: Yep, this is the new structure mentioned on dash PR, we need some tests

ureport/polls/models.py
```diff
@@ -346,7 +346,7 @@ def get_results(self, segment=None):
         if cached_value:
             return cached_value['results']
 
-        if segment and segment.get('location', "") == "District":
+        if segment and segment.get('location', "") in ('District', 'Ward'):
```
norkans7: unit test for this

norkans7:
This needs rapidpro to have the support for third level first so it will wait for that first

musamusa:
@norkans7 I'm  updating PR now. Also third level support added to rapidpro. will buzz you ASAP when done with the updates.

nicpottier:
@norkans7 plz review again.

norkans7:
I realize this will not work as we changed the way we calculate the stats for reporters in #81 and #82 to use database triggers

And still in development and merging soon poll results #127

This need to follow the same logic for ward, probably after the poll results are merged

PR #114, opened 2016-01-29
Fix reporters count type

norkans7:
- make contact org and uuid unique together
- fix reporters count type field
- Update smartmin and rapidpro client


norkans7:
Included in #117

PR #116, opened 2016-02-02
remove Uganda from old sites settings

norkans7:
- Removes Uganda from the Old site settings config
- Update tests


nicpottier:
Yarp.

norkans7:
@nicpottier I added the subdomain ordering here as well , please review

PR #117, opened 2016-02-02
Poll responses

norkans7:
In addition to #115 and #114 

- Add poll flow base language
- Add response category
- migrations to populate poll responses from rapidpro flow definition


nicpottier:
Looks good.

PR #118, opened 2016-02-04
Add a way to deactivate the contacts that opted out on rapidpro

norkans7:
- Added a field is_active which defaults to True and turned False if the contact opted out on Rapidpro
- we keep that contact so if we fetch old poll result we can match to it
- Update trigger to adjust the contact for the is_active field, if False remove one from all reporters counts
otherwise add one

### DIFF
--- a/ureport/contacts/migrations/0010_contact_triggers.py
+++ b/ureport/contacts/migrations/0010_contact_triggers.py
@@ -1,0 +1,163 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+# language=SQL
+TRIGGER_SQL = """
+-----------------------------------------------------------------------------
+-- Increment or decrement the reporters count on counter of given type
+-----------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION
+  ureport_insert_reporters_counter(_org_id INT, _type VARCHAR, _count INT)
+RETURNS VOID AS $$
+BEGIN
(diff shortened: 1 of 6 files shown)


ureport/contacts/migrations/0010_contact_triggers.py
```diff
@@ -0,0 +1,163 @@
...
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION ureport_adjust_counter_for_contact(_new_contact contacts_contact, _old_contact contacts_contact)
+RETURNS VOID AS $$
+BEGIN
+  -- no org id, decrement all reporters counters for its previous values
+  IF _new_contact.org_id IS NULL THEN
+    PERFORM ureport_increment_counter_for_contact(_old_contact, FALSE);
+
+  ELSIF _new_contact.is_active != _old_contact.is_active THEN
```
norkans7: This is the only change in trigger sql to handle is_active field update

norkans7:
@nicpottier please review

norkans7:
This will need a refetch of all contacts to be accurate, 
and I am open to other suggestions

norkans7:
This relied on that now Rapidpro updates modified_on when a contact is added or removed from a group,

nicpottier:
Looks good.

norkans7:
@nicpottier @rowanseymour Also switched to use v2 for contacts fetch, please have a second review

nicpottier:
Yep.

norkans7:
functionality included in #125

PR #119, opened 2016-02-05
use api v2 to get contacts

norkans7:


ureport/contacts/models.py
```diff
@@ -247,27 +248,19 @@ def fetch_contacts(cls, org, after=None):
             # consider the after year 2013
             after = json_date_to_datetime("2013-01-01T00:00:00.000")
 
-        while before > after:
-            pager = temba_client.pager()
-            api_contacts = temba_client.get_contacts(before=before, after=after, pager=pager)
+        api_contacts_query = temba_client_2.get_contacts(before=before, after=after)
+        api_contacts = api_contacts_query.all()
```
rowanseymour: Better to use `.iterfetches(retry_on_rate_exceed=True)` here instead of `.all()` and update the contacts in batches. Otherwise all will try to fetch every single modified contact which for an org like Nigeria will take forever and require a lot of memory. See https://github.com/rapidpro/casepro/blob/5bb79cd03fdb45ec228a334c8592b34e98e3a911/casepro/contacts/sync.py#L48-L48

rowanseymour: `retry_on_rate_exceed` also tells the client to sleep and retry if you hit the rate limit. Without it the client just returns an exception.

ureport/contacts/models.py
```diff
@@ -247,27 +248,19 @@ def fetch_contacts(cls, org, after=None):
             # consider the after year 2013
             after = json_date_to_datetime("2013-01-01T00:00:00.000")
```
rowanseymour: This is an odd way to initialise a datetime variable

norkans7: this can be removed now before it made sense with the while loop

ureport/contacts/models.py
```diff
@@ -243,31 +244,18 @@ def fetch_contacts(cls, org, after=None):
...
-                if i == last_contact_index:
-                    before = contact.modified_on
-
-                if group_uuid in contact.groups:
-                    cls.update_or_create_from_temba(org, contact)
-                    seen_uuids.append(contact.uuid)
-                else:
-                    removed_uuids.append(contact.uuid)
+        for contact in api_contacts_query.iterfetches(retry_on_rate_exceed=True):
+            if group_uuid in contact.groups:
```
rowanseymour: Groups in the API v2 model is now a list of dicts with uuid and name of each group

PR #125, opened 2016-03-07
Use syncers

norkans7:
- Update dash
- Update rapidpro client
- Use dash syncers for contacts fetch from rapidpro

### DIFF
--- a/ureport/backend/rapidpro.py
+++ b/ureport/backend/rapidpro.py
@@ -1,0 +1,210 @@
+from __future__ import unicode_literals
+
+import json
+
+from dash.utils.sync import BaseSyncer, sync_local_to_set, sync_local_to_changes
+
+from ureport.contacts.models import ContactField, Contact
+from ureport.locations.models import Boundary
+from . import BaseBackend
+
+from ureport.utils import is_dict_equal
+
+
+class FieldSyncer(BaseSyncer):
+    """
+    Syncer for contact fields
+    """
+    model = ContactField
+    local_id_attr = 'key'
+    remote_id_attr = 'key'
+
+    def local_kwargs(self, org, remote):
+        return {
+            'org': org,
+            'key': remote.key,
+            'label': remote.label,
+            'value_type': self.model.TEMBA_TYPES.get(remote.value_type, self.model.TYPE_TEXT)
+        }
+
+    def update_required(self, local, remote):
+        return local.label != remote.label or local.value_type != self.model.TEMBA_TYPES.get(remote.value_type)
+
+
+class BoundarySyncer(BaseSyncer):
+    """
+    syncer for location boundaries
+    """
+    model = Boundary
+    local_id_attr = 'osm_id'
+    remote_id_attr = 'boundary'
+
+    def local_kwargs(self, org, remote):
+        geometry = json.dumps(dict(type=remote.geometry.type, coordinates=remote.geometry.coordinates))
+
+        parent = Boundary.objects.filter(osm_id__iexact=remote.parent, org=org).first()
+
+        return {
+            'org': org,
+            'geometry': geometry,
+            'parent': parent,
+            'level': remote.level,
+            'name': remote.name,
+            'osm_id': remote.boundary
+        }
+
+    def update_required(self, local, remote):
+        if local.name != remote.name:
+            return True
+
+        if local.level != remote.level:
+            return True
+
+        if remote.parent:
+            if not local.parent:
+                return True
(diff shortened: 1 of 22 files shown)


ureport/backend/rapidpro.py
```diff
@@ -0,0 +1,210 @@
...
+            'gender': gender,
+            'born': born,
+            'occupation': occupation,
+            'registered_on': registered_on,
+            'state': state,
+            'district': district
+        }
+
+    def update_required(self, local, remote):
+        remote_kwargs = self.local_kwargs(local.org, remote)
```
rowanseymour: Hmm calling `local_kwargs` again here isn't ideal - I think we should change `update_required` so that the kwargs are passed to it

ureport/backend/rapidpro.py
```diff
@@ -0,0 +1,210 @@
...
+        state_field = org.get_config('state_label')
+        if state_field:
+            if org.get_config('is_global'):
+                state_name = remote.fields.get(self.model.find_contact_field_key(org, state_field), None)
+                if state_name:
+                    state = state_name
+
+            else:
+                state_name = remote.fields.get(self.model.find_contact_field_key(org, state_field), None)
+                state_boundary = Boundary.objects.filter(org=org, level=1, name__iexact=state_name).first()
```
rowanseymour: Is is possible to store these on the syncer rather than reload for each contact? Remember you should be trying to minimizing the number of db hits you make for each contact because sometimes there will 1.4 million of them.

ureport/polls/migrations/0020_auto_20150508_1210.py
```diff
@@ -1,8 +1,44 @@
...
+import time
+from django.conf import settings
 from django.db import models, migrations
-from dash.api import API
+
+
+def fetch_flows(filter=None):
+    start = time.time()
+
+    next = '%s/api/v1/flows.json' % settings.SITE_API_HOST
```
rowanseymour: Why not use the v1 client here?

ureport/settings_common.py
```diff
@@ -488,11 +492,11 @@
...
         "relative": True,
     },
-    "fetch_contacts": {
-        "task": "contacts.fetch_contacts_task",
-        "schedule": timedelta(minutes=10),
-        "relative": True,
-     },
+    'contact-pull': {
+        'task': 'dash.orgs.tasks.trigger_org_task',
+        'schedule': crontab(minute=[0, 10, 20, 30, 40, 50]),
```
rowanseymour: Btw the reason I used this crontab in casepro was to stagger the contact and message pull tasks so they don't launch at the same time

ureport/utils/__init__.py
```diff
@@ -44,6 +46,22 @@ def json_date_to_datetime(date_str):
     return datetime.strptime(date_str, iso_format).replace(tzinfo=pytz.utc)
 
 
+def is_dict_equal(d1, d2, keys=None, ignore_none_values=True):
```
rowanseymour: Lets move this function to `dash.utils` if you plan to use it

ureport/backend/rapidpro.py
norkans7: Yes, right

ureport/backend/rapidpro.py
norkans7: sure

ureport/polls/migrations/0020_auto_20150508_1210.py
norkans7: this is an old migration I am keeping happy to run, I think it is okay to use the client as well and only make flows be a list of the data were only interested in.

ureport/utils/__init__.py
norkans7: sure

ureport/settings_common.py
norkans7: Yep, I was planning for the same once we add runs fetch or poll results, And I was thinking this will make easy to see the task were not scheduled when expected too

ureport/polls/migrations/0020_auto_20150508_1210.py
norkans7: I think now v1 client cannot work since it does not return the ruleset ids which we need in this migration

rowanseymour: That's fine - i didn't notice this was just an old migration

nicpottier:
Looks reasonable to me.. excited to get onto poll runs!

PR #127, opened 2016-03-12
Use Runs API for poll results part 1

norkans7:
Part 1 to fetch poll results from runs endpoint


ureport/backend/rapidpro.py
```diff
@@ -251,3 +257,83 @@ def pull_contacts(self, org, modified_after, modified_before, progress_callback=
...
+
+                for temba_step in temba_run.steps:
+                    ruleset_uuid = temba_step.node
+                    category = temba_step.category
+                    text = temba_step.text
+
+                    if PollResult.objects.filter(org=org, flow=flow_uuid,
+                                                 ruleset=ruleset_uuid, contact=contact_uuid).exists():
+
+                        if not PollResult.objects.filter(org=org, flow=flow_uuid, ruleset=ruleset_uuid,
```
rowanseymour: If you actually fetch the object on line 301, you don't have fetch it again here

norkans7: I am using exists()

I think of changing this to 
```
if PollResult.objects.filter(org=org, flow=flow_uuid, ruleset=ruleset_uuid,
                                                  contact=contact_uuid).update(category=category, text=text,
                                                                               state=state, district=district,
                                                                               completed=completed):
                        num_updated += 1

                    else:
                        PollResult.objects.create(org=org, flow=flow_uuid, ruleset=ruleset_uuid, contact=contact_uuid,
                                                  category=category, text=text, state=state, date=temba_step.arrived_on,
                                                  district=district, completed=completed)
```

ureport/backend/rapidpro.py
```diff
@@ -251,3 +257,83 @@ def pull_contacts(self, org, modified_after, modified_before, progress_callback=
...
+                                                         contact=contact_uuid,category=category, text=text,
+                                                         state=state, district=district, completed=completed).exists():
+
+                            PollResult.objects.filter(org=org, flow=flow_uuid,
+                                                      ruleset=ruleset_uuid,
+                                                      contact=contact_uuid).update(category=category, text=text,
+                                                                                   state=state, district=district,
+                                                                                   completed=completed)
```
nicpottier: Same here, could just do a save() or query by id and do an update.

ureport/backend/rapidpro.py
```diff
@@ -251,3 +257,83 @@ def pull_contacts(self, org, modified_after, modified_before, progress_callback=
...
+
+        start = time.time()
+        print "Start fetching runs for org #%d" % org.pk
+
+        # ignore the TaskState time and use the time we stored in redis
+        now = timezone.now()
+        after = cache.get(PollResult.POLL_RESULTS_LAST_PULL_CACHE_KEY % (org.pk, poll.pk), None)
+
+        poll_runs_query = client.get_runs(flow=poll.flow_uuid, responded=True, after=after, before=now)
+        fetches = poll_runs_query.iterfetches(retry_on_rate_exceed=True)
```
rowanseymour: Poll results can only be modified by this syncing process which makes things easier to optimize since you don't have to worry about another process trying to make changes to the same objects at the same thing. For a start you could loop on each fetch of 250 results and load the existing results from the db in one query for all of that fetch.

rowanseymour: i.e. down on line ~284 load all `PollResult` objects that match the flow and contacts in `fetch`

ureport/backend/rapidpro.py
rowanseymour: Downside of that would be that you're overwriting all the run steps even if they haven't changed - and db writes are usually the slow part of these syncing routines.

ureport/polls/migrations/0032_install_poll_results_count_triggers.py
```diff
@@ -0,0 +1,121 @@
...
+BEGIN
+  INSERT INTO polls_pollresultscounter("org_id", "ruleset", "type", "count") VALUES(_org_id, _ruleset, _type, _count);
+  PERFORM ureport_may_be_squash_resultscounters(_org_id, _ruleset, _type);
+END;
+$$ LANGUAGE plpgsql;
+-----------------------------------------------------------------------------
+-- Every 100 inserts or so this will squash the counters by gathering
+-----------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION
+  ureport_may_be_squash_resultscounters(_org_id INT, _ruleset CHAR(36), _type VARCHAR)
```
nicpottier: ```maybe```

nicpottier:
Not seeing the migration to drop the index on ```org``` so it always uses the compound index.

norkans7:
0035

nicpottier:
Oh nice gotcha.

All looks pretty straightforward to me..

PR #135, opened 2016-04-08
add a way pull poll results for many polls simultaneously

norkans7:
Currently org fetch on poll at time and use one worker only

This adds a way to trigger many tasks one for each poll
so we can use all the workers to backfill the poll results at the same time.

### DIFF
--- a/ureport/polls/models.py
+++ b/ureport/polls/models.py
@@ -100,6 +100,13 @@ class Poll(SmartModel):
             question.fetch_results(dict(location='State'))
 
     @classmethod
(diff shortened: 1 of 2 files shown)


norkans7:
@nicpottier Please review

nicpottier:
Are you sure this is going to be faster? These tasks seem to be DB bound on the U-Report side, so just adding more tasks isn't going to make things finish any faster as they will all be running at half the speed no?

norkans7:
sawa

PR #137, opened 2016-04-14
squash poll results counter every 10000 instead of 100 to speed up th…

norkans7:
…e sync

### DIFF
--- a/ureport/polls/migrations/0037_install_poll_results_count_triggers.py
+++ b/ureport/polls/migrations/0037_install_poll_results_count_triggers.py
@@ -18,7 +18,7 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 -----------------------------------------------------------------------------
--- Every 100 inserts or so this will squash the counters by gathering
+-- Every 10000 inserts or so this will squash the counters by gathering
 -----------------------------------------------------------------------------
 CREATE OR REPLACE FUNCTION
   ureport_maybe_squash_resultscounters(_org_id INT, _ruleset CHAR(36), _type VARCHAR)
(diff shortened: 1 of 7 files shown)


ureport/polls/migrations/0037_install_poll_results_count_triggers.py
```diff
@@ -0,0 +1,120 @@
...
+END;
+$$ LANGUAGE plpgsql;
+-----------------------------------------------------------------------------
+-- Every 10000 inserts or so this will squash the counters by gathering
+-----------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION
+  ureport_maybe_squash_resultscounters(_org_id INT, _ruleset CHAR(36), _type VARCHAR)
+RETURNS VOID AS $$
+BEGIN
+  IF RANDOM() < .0001 THEN
```
norkans7: change to squash every 10000 inserts instead of 100

rowanseymour:
Have you thought about putting trigger sql in separate .sql files like rapidpro and casepro? Does make it a little easier to keep track of trigger stuff.

Can't see from diff but is only thing to change here the squash frequency?

norkans7:
Yep the frequency is the only thing changed from https://github.com/rapidpro/ureport/blob/master/ureport/polls/migrations/0032_install_poll_results_count_triggers.py#L27

I can change to use .sql files as well

nicpottier:
This is the only change you've found to speed up the writes? What are the timings on the various pieces of the triggers?

norkans7:
Please review

nicpottier:
Seems good, curious to see what the performance is of squashing 250 runs at a time..

nicpottier:
do it!

PR #139, opened 2016-04-19
Add methods to rebuild the counts from the poll results

norkans7:


ureport/polls/models.py
```diff
@@ -513,6 +515,96 @@ class PollResult(models.Model):
...
+
+        start = time.time()
+
+        flow_org_pairs = list(Poll.objects.all().values('org_id', 'flow_uuid').order_by('org_id', 'flow_uuid').distinct('org_id', 'flow_uuid'))
+
+        for flow_org_pair in flow_org_pairs:
+            pair_start = time.time()
+            org_id = flow_org_pair['org_id']
+            flow = flow_org_pair['flow_uuid']
+            pair_results = list(PollResult.objects.filter(org_id=org_id, flow=flow).order_by('org_id', 'flow'))
```
nicpottier: This is going to be huge if you get all the objects.. probably want to do a value_list of the ids here instead and look up the actual objects per batch.

ureport/polls/models.py
```diff
@@ -513,6 +515,96 @@ class PollResult(models.Model):
...
+
+        if district and category:
+            update_counters.append(PollResultsCounter.get_or_create(org_id=org_id, ruleset=ruleset,
+                                                                    type='ruleset:%s:category:%s:district:%s' % (ruleset, category, district)).id)
+        elif district:
+            update_counters.append(PollResultsCounter.get_or_create(org_id=org_id, ruleset=ruleset,
+                                                                    type='ruleset:%s:nocategory:district:%s' % (ruleset, district)).id)
+
+        PollResultsCounter.objects.filter(id__in=update_counters).update(count=F('count') + 1)
+
```
nicpottier: Since get_or_create() can return more than one, aren't you incrementing all of them by 1 here instead of only incrementing one of them?

ureport/polls/models.py
```diff
@@ -513,6 +515,96 @@ class PollResult(models.Model):
 
     district = models.CharField(max_length=255, null=True)
 
+    @classmethod
+    def rebuild_counts(cls):
```
nicpottier: Seems like being able to rebuild the counts for just one flow would likely be useful

ureport/polls/models.py
norkans7: this is a class method I added not the one on the manager, Yes, I make sure I increment only one object

ureport/polls/models.py
norkans7: yep

ureport/polls/models.py
norkans7: I had assumed and reinforced the counter table was TRUNCATED but I can change this instead to first delete the counters and rebuild them form results, that's much better

ureport/polls/models.py
```diff
@@ -531,6 +623,16 @@ class PollResultsCounter(models.Model):
     count = models.IntegerField(default=0, help_text=_("Number of items with this counter"))
 
     @classmethod
+    def get_or_create(cls, org_id, ruleset, type):
+        try:
+            counter, created = PollResultsCounter.objects.get_or_create(org_id=org_id, ruleset=ruleset, type=type)
+        except MultipleObjectsReturned:
+            counter = PollResultsCounter.objects.filter(org_id=org_id, ruleset=ruleset, type=type)
+
```
nicpottier: Doesn't this return more than one?

norkans7: my bad I missed .first()

nicpottier:
shipit

nicpottier:
shipit

PR #141, opened 2016-04-20
feature/eha/ward level implementaion

musamusa:
## The Following was updated ##

* Polls Location map was updated to show level 3 boundaries (ward)
* Polls map Scale was updated to indicate third level
* U-Reporters  Location map was updated to show level 3 boundaries (ward)
* U-Reporters map Scale was updated to indicate third level

selection_031
selection_032
selection_033


musamusa:
@norkans7  please review

static/coffee/locations.coffee
```diff
@@ -128,16 +133,27 @@ initMap = (id, geojson, question, districtLabel) ->
...
+    else if boundary and boundary.level == DISTRICT_LEVEL
+      segment = {location:"Ward", parent:boundaryId}
+      overallResults = boundaryResults[boundaryId]
     else
       segment = {location:"District", parent:boundaryId}
       overallResults = boundaryResults[boundaryId]
 
     $.ajax({url:'/pollquestion/' + question + '/results/?segment=' + encodeURIComponent(JSON.stringify(segment)), dataType: "json"}).done (data) ->
+      data = data || []
+      if data.length == 0
```
norkans7: I think this is a bad experience to the site visitor because he is not aware the the map will zoom back if there is no data. We just show the location with grey(zero info)

static/coffee/ureporters.coffee
```diff
@@ -104,25 +106,39 @@ initMap = (id, geojson, ajaxUrl, districtLabel, colorsList=[]) ->
...
     if not boundaryId
       segment = {location:"State"}
+    else if boundary and boundary.level == DISTRICT_LEVEL
+      segment = {location:"Ward", parent:boundaryId}
     else
       segment = {location:"District", parent:boundaryId}
 
     $.ajax({url: ajaxUrl + '&segment=' + encodeURIComponent(JSON.stringify(segment)), dataType: "json"}).done (data) ->
+      data = data || []
+      if data.length == 0
```
norkans7: bad experience same here

static/less/ureport2.less
```diff
@@ -1839,33 +1857,49 @@ ul.poll-filters, .featured-stories-filters, ul.stories-filters {
   .state-level {
     position: absolute;
     top: 30px;
-    right: 8px;
```
norkans7: using right attribute is the right thing here some translations could be shown in part

ureport/backend/rapidpro.py
```diff
@@ -221,6 +237,7 @@ def update_required(self, local, remote, local_kwargs):
         update = update or local.occupation != local_kwargs['occupation']
         update = update or local.registered_on != local_kwargs['registered_on']
         update = update or local.state != local_kwargs['state'] or local.district != local_kwargs['district']
+
```
norkans7: we need to check the ward has changed and make ```update = True```

or local ward is not the same as the one from remote then update must be True

ureport/locations/models.py
```diff
@@ -8,10 +8,15 @@
...
 
+
 class Boundary(models.Model):
     """
     Corresponds to a RapidPro AdminBoundary
     """
+    COUNTRY_LEVEL = 0
+    STATE_LEVEL = 1
+    DISTRICT_LEVEL = 2
+    WARD_LEVEL = 3
```
norkans7: Nice for the constants

ureport/polls/models.py
```diff
@@ -426,7 +426,7 @@ def get_results(self, segment=None):
         if cached_value:
             return cached_value['results']
 
-        if segment and segment.get('location', "") == "District":
+        if segment and segment.get('location', "") in ("District", "Ward"):
```
norkans7: this way of fetching data is going to be removed soon, That's why we have ```PollResult``` model

ureport/utils/__init__.py
```diff
@@ -487,20 +489,25 @@ def get_ureporters_locations_stats(org, segment):
...
-        boundaries = Boundary.objects.filter(org=org, level=boundary_top_level).values('osm_id', 'name').order_by('osm_id')
+        boundary_top_level = Boundary.COUNTRY_LEVEL if org.get_config('is_global') else Boundary.STATE_LEVEL
+        boundaries = Boundary.objects.filter(org=org, level=boundary_top_level).values('osm_id', 'name')\
+            .order_by('osm_id')
         location_counts = {k[6:]: v for k, v in org_contacts_counts.iteritems() if k.startswith('state')}
-
+    elif field_type == 'ward':
+        boundaries = Boundary.objects.filter(org=org, level=Boundary.WARD_LEVEL,
+                                             parent__osm_id__iexact=parent).values('osm_id', 'name').order_by('osm_id')
+        location_counts = {k[9:]: v for k, v in org_contacts_counts.iteritems() if k.startswith('ward')}
```
norkans7: the number 9 there is the length of district + 1 so here it should be 5 as (len('ward') + 1)

the reason being the we use Postgres triggers to keep the counts; so that we query the counter table (```PollResultCounter``` model) which is faster that querying the results table (```PollResult``` model)

norkans7:
The important thing missing is that you need to add a migration to update the postgres routines used by the triggers to also make counts of the ward stats

something like https://github.com/rapidpro/ureport/blob/master/ureport/sql/polls_0037.sql
used in https://github.com/rapidpro/ureport/blob/master/ureport/polls/migrations/0037_install_poll_results_count_triggers.py

and the same for contacts and reporterscounter

musamusa:
@norkans7 updated

ureport/sql/polls_0037.sql
```diff
@@ -46,6 +46,13 @@ BEGIN
       -- Maps counts without replies; contacts from this district who did not reply
       PERFORM ureport_insert_results_counter(_poll_result.org_id, _poll_result.ruleset, CONCAT('ruleset:', LOWER(_poll_result.ruleset), ':nocategory:', 'district:', UPPER(_poll_result.district)), _count);
     END IF;
+    IF _poll_result.ward IS NOT NULL AND _poll_result.category IS NOT NULL THEN
```
norkans7: This is not quite right, you need to copy this file into a new one and name it as the migration that runs it if the migrations is 0038_... the sql file will be polls_0038 and in the migration you InstallSQL('poll_0038')

static/coffee/locations.coffee
```diff
@@ -238,6 +231,12 @@ initMap = (id, geojson, question, districtLabel) ->
         boundaryUrl += boundaryId + '/'
 
       $.ajax({url:boundaryUrl, dataType: "json"}).done (data) ->
+        # added to reset boundary when district has no wards
```
norkans7: This too will still be bad to the visitor; the way we had made this is to prevent zooming if the location level is not configured look for allowDistrictZoom

norkans7:
@xkmato the migrations needs to be done for contacts and reporterscounters

norkans7:
https://github.com/rapidpro/ureport/pull/140 has the sql file for contacts that you can start with just waiting on your decisio on #79 to remove the conflicting commit on this PR before merging

static/coffee/locations.coffee
musamusa: @norkans7 this was added to prevent loading of empty boundary data for districts that don't have ward level data. If I remove that check it will throw  ```TypeError: this._northEast is undefined```. any ideas on how to prevent that without that check? so far it worked well without breaking anything.

musamusa:
@norkans7 another update

static/coffee/locations.coffee
norkans7: I think we do not configure that level on the org if we do not have enough data for the country on a level, then you need to add a flag allWardZoom like https://github.com/rapidpro/ureport/pull/141/files#diff-93afaaac4631d6e1b07a129bbcb538c1R7

which we check before zooming to the level

musamusa: @norkans7 I understand what you mean, but my point here is that currently in Nigeria, we only have ward level boundaries for 10 states out of 36 states, so if allWardZoom is enabled it will throw  ```TypeError: this._northEast is undefined``` for states without boundaries. I can remove the check but that will mean visitors will have to reload page to view map if the error occurs.

norkans7: Ok, keep this and I think it is still good to have the other way as well since this will have a delay for the ajax call. so we use the zoomDisabled if the level is not configured

norkans7:
the same for as pollresults counter you need to add a migration for ward for reporters #143 has a start sql file

And also some unit tests to make sure the counts are right and for those features you added

musamusa:
@norkans7 updated

norkans7:
Hey @musamusa, I do not see yet unit tests, 
and since we are working mostly in the same land there is a high chance to step on each other and making this more complicated to merge.

I think it would be a good approach to break this in different PRs the first dealing with adding ward and its migrations and tests on the backend, and the other we go into the details of adding the data to the frontend(public site)

musamusa:
@norkans7 I updated the current test to cover my additions which improved the current test coverage, also the agreement was also not to reduce current coverage.

norkans7:
Right, however we need new tests to prevent future changes to break these additions

ureport/contacts/tests.py
```diff
@@ -74,25 +74,33 @@ def setUp(self):
...
                                                geometry='{"foo":"bar-country"}')
-        self.state = Boundary.objects.create(org=self.nigeria, osm_id="R-LAGOS", name="Lagos", level=1,
+        self.state = Boundary.objects.create(org=self.nigeria, osm_id="R-LAGOS", name="Lagos",
+                                             level=Boundary.STATE_LEVEL,
                                              parent=self.country, geometry='{"foo":"bar-state"}')
-        self.district = Boundary.objects.create(org=self.nigeria, osm_id="R-OYO", name="Oyo", level=2,
+        self.district = Boundary.objects.create(org=self.nigeria, osm_id="R-OYO", name="Oyo",
+                                                level=Boundary.DISTRICT_LEVEL,
                                                 parent=self.state, geometry='{"foo":"bar-state"}')
+        self.district = Boundary.objects.create(org=self.nigeria, osm_id="R-IKEJA", name="Ikeja",
```
norkans7: self.district here was declared before, not quite right

ureport/contacts/tests.py
```diff
@@ -74,25 +74,33 @@ def setUp(self):
...
+                                                level=Boundary.WARD_LEVEL,
+                                                parent=self.district, geometry='{"foo":"bar-ward"}')
 
         self.registration_date = ContactField.objects.create(org=self.nigeria, key='registration_date',
                                                              label='Registration Date', value_type='T')
 
         self.state_field = ContactField.objects.create(org=self.nigeria, key='state', label='State', value_type='S')
         self.district_field = ContactField.objects.create(org=self.nigeria, key='lga', label='LGA', value_type='D')
+        self.district_field = ContactField.objects.create(org=self.nigeria, key='ward', label='Ward', value_type='W')
```
norkans7: same here

musamusa:
@norkans7 added one test in contacts and fixed duplicate declarations

ureport/contacts/tests.py
```diff
@@ -246,6 +246,19 @@ def test_fetch_contacts(self):
                     Contact.fetch_contacts(self.nigeria, after=datetime(2014, 12, 01, 22, 34, 36, 123000, pytz.utc))
                     self.assertTrue('000-001' in seen_uuids)
 
+    def test_contact_ward_field(self):
+        temba_contact = TembaContact.create(uuid='C-0011', name="Jan", urns=['tel:123'],
+                                            groups=['G-001', 'G-007'],
+                                            fields={'registration_date': '2014-01-02T03:04:05.000000Z', 'state':'Lagos',
+                                                    'lga': '', 'ward': 'Ikeja', 'occupation': 'Student', 'born': '1990',
```
norkans7: Yes, this is progress, this is one case where we ignore the ward since we do not know the district

we need more tests for all cases, when the ward is recognized as well for example
so you need to test the poll results as well, and the counts calculated by Postgres triggers

Again consider breaking the PR into many PR

musamusa:
@norkans7 PR broken into frontend and backend as requested both at #145 and #147

norkans7:
Yep, let's close this and consider #145 and #147

PR #142, opened 2016-04-20
Api poll data

norkans7:


norkans7:
added both on the previous PR #140

PR #144, opened 2016-04-22
use python to calculate to aggregate the counts and insert them at once

norkans7:


norkans7:
decrease the time to squash a poll from 10k secs to 550secs
another poll from 2500s to 121s

ureport/polls/models.py
```diff
@@ -567,19 +568,88 @@ def rebuild_counts_for_poll(cls, poll_id):
                 print "Results query time for pair %s, %s took %ds" % (org_id, flow, time.time() - start)
 
                 processed_results = 0
+                counters_dict = dict()
+                key_ruleset_map = dict()
+                key_org_id_map = dict()
```
nicpottier: Pretty sure you can do something like ```defaultdict(int)``` here and not have to test for membership before incrementing..

ureport/polls/models.py
```diff
@@ -567,19 +568,88 @@ def rebuild_counts_for_poll(cls, poll_id):
...
                     for result in poll_results:
-                        result.create_counters()
+                        gen_counters = result.generate_counters()
+                        for key in gen_counters.keys():
+                            if key in counters_dict:
+                                counters_dict[key] += gen_counters[key]
+                            else:
+                                counters_dict[key] = gen_counters[key]
+                            key_ruleset_map[key] = result.ruleset
+                            key_org_id_map[key] = result.org_id
```
nicpottier: Wonder if it makes sense to just use a tuple of (org_id, ruleset, key) as the key in your counter maps instead of this. Might be a bit simpler overall.

nicpottier:
apart from those tweaks looks good

ureport/polls/models.py
norkans7: 👍

ureport/polls/models.py
norkans7: okay,

nicpottier:
noice

ureport/polls/models.py
```diff
@@ -567,19 +569,81 @@ def rebuild_counts_for_poll(cls, poll_id):
...
+                counters_to_insert = []
+                for counter_tuple in counters_dict.keys():
+                    org_id, ruleset, counter_type = counter_tuple
+                    count = counters_dict[counter_tuple]
+                    counters_to_insert.append(PollResultsCounter(org_id=org_id, ruleset=ruleset, type=counter_type,
+                                                                 count=count))
+
+                PollResultsCounter.objects.bulk_create(counters_to_insert)
+                # now squash the counters
+                PollResultsCounter.squash_counts(from_zero=True)
```
norkans7: squash from zero so we lookup duplicates only, this saves a lot of time since we already sure the counters we insert are unique from python

PR #153, opened 2016-05-10
Fix API path in docs swagger

teehamaral:
With this fix you can make requests within the API documentation, which is currently repeating the prefix `/api/v1/` example: `http://ureport.in:80/api/v1/api/v1/polls/org/1/`.


nicpottier:
Yep and thanks!

PR #133, opened 2016-04-06
V2 poll results Part 2

norkans7:
This will use the backfilled results on the public site

PR #154, opened 2016-05-11
Use the fetched data from runs to the public page

norkans7:
this is sort of reapplying #133 which was reverted in #152 to use the poll results on the public site after the backfill is finished


nicpottier:
go for it

nicpottier:
looks sane, have you tested on staging?

PR #157, opened 2016-05-12
poll results insert on every 250 runs fetch to manage memory usage

norkans7:
Use less memory


norkans7:
@nicpottier please review

nicpottier:
Ship!

PR #169, opened 2016-06-02
update deps to support python 2.7.11

norkans7:


nicpottier:
Yep.. no `BROKER_BACKEND` in settings_common?

norkans7:
It is set to 'redis'

PR #171, opened 2016-06-03
every 10 min check the if we need to update the cache for question re…

norkans7:
…sults

This will prevent the timeouts on poll pages not visited often

### DIFF
--- a/ureport/polls/models.py
+++ b/ureport/polls/models.py
@@ -488,7 +488,7 @@ class PollQuestion(SmartModel):
     """
 
     POLL_QUESTION_RESULTS_CACHE_KEY = "org:%d:poll:%d:question_results:%d"
-    POLL_QUESTION_RESULTS_CACHE_TIMEOUT = 60 * 5
+    POLL_QUESTION_RESULTS_CACHE_TIMEOUT = 60 * 12
 
     poll = models.ForeignKey(Poll, related_name='questions',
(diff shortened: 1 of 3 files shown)


nicpottier:
Is this actually resetting the expiration date?
Also, why expire at all?

norkans7:
```get_results``` is the same method that tries to get from the cache if the cache misses it continue to the calculations and I think recalculating all the segmentations very often will be a big load

nicpottier:
So what you have here isn't at all guaranteeing that the cache is alive, there's still the possibility of a gap between when it expires and this job runs and then the page will time out.

You either need to set a longer timeout and recalculate these (and set a new timeout) before it expires or just not expire things, otherwise there will still be times when people hit pages with no cached data.

norkans7:
Here the cache is always alive for the page to load, but there other times like segmenting by district we use the same method.

Or may be we do not expire for these cases only(no segment, segment by state)? and soon by gender and age

nicpottier:
We need to be able to load the page instantly, so seems like you need the main view, segment by state, gender, age always ready.

Wasn't the idea to have the results counted up so these were fast queries? Storing all of these in Redis is not likely going to work as Redis needs to keep everything in memory all the time.

norkans7:
I think the failing case is he open ended question which have to go to the DB to aggregate the frequent words from msg texts

PR #172, opened 2016-06-06
Optimize open ended

norkans7:
- Move the open ended word counts for word cloud from SQL to python

- Remove the tasks added previously to update the cache

### DIFF
--- a/ureport/polls/models.py
+++ b/ureport/polls/models.py
@@ -3,6 +3,7 @@ import json
 from collections import defaultdict
 
 import pytz
+import re
 from django.contrib.auth.models import User
 from django.db import models, connection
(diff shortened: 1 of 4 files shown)


norkans7:
@nicpottier please review

nicpottier:
Doesn't seem like the python counting is going to be faster than SQL here.. what happens with really big results?

norkans7:
I saw were including the pollresults with text null(1.2M), I started by filtering those out (60k) the DB query reduce from 4s to 1s, and then thought I can move the calculation on python side and the query will  be around 90ms

I also run a vacuum since the query was using a bitmap scan after the index scan.

PR #175, opened 2016-06-07
Change deprecated redis cache

norkans7:


ureport/polls/tests.py
```diff
@@ -1111,9 +1111,9 @@ def test_delete_poll_results_counter(self):
                                   ruleset=poll_question.ruleset_uuid, date=timezone.now(),
                                   contact='contact-uuid', completed=False)
 
-        with self.settings(CACHES={'default': {'BACKEND': 'redis_cache.cache.RedisCache',
+        with self.settings(CACHES={'default': {'BACKEND': 'django_redis.cache.RedisCache',
```
rowanseymour: Why do these settings have to be overridden on a per test basis?

norkans7: The default is the Dummy cache, so we are overriding whenever we need to consider the cached value

rowanseymour: Any reason you're using dummy cache? Don't we usually just use redis db 15?

norkans7: yep, I can update and use always redis, and adjust the place we need to ignore the cache

norkans7:
please review

nicpottier:
Was caching not working? What's the functional change here?

norkans7:
Nothing changed, I just founf redis_cache was deprecated and by using django_redis, @rowanseymour  catched we were repeating the cache settings unnecessary, so I went ahead and made that adjustments

PR #176, opened 2016-06-08
Calculate results

norkans7:
- Using the org tasks is having issue since the redis locks are not released if the task is killed by deploy
and that prevent the tasks to run again.

- more test adjustments to not consider the caches


nicpottier:
yep

PR #192, opened 2016-06-28
Fix conditional twitter handle

teehamaral:
Fixing the conditional that checks whether the organization has a twitter handle.

### DIFF
--- a/templates/public/join_engage.haml
+++ b/templates/public/join_engage.haml
@@ -134,5 +134,5 @@
 - block extra-script
   {{ block.super }}
   -if not is_iorg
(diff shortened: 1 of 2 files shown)


templates/public/join_engage.haml
```diff
@@ -134,5 +134,5 @@
 - block extra-script
   {{ block.super }}
   -if not is_iorg
-    if org|config:"twitter_handle"
+    -if org|config:"twitter_handle"
```
norkans7: use one if block here then instead on nested ifs

norkans7:
Yep

PR #202, opened 2016-08-09
Ignore texts that looks like links for open ended questions

norkans7:


norkans7:
Please review

nicpottier:
Did you try running this query on prod? That ILIKE might be pretty expensive

norkans7:
Okay I can try that first

norkans7:
I see both are in the same range

nicpottier:
then doit!

PR #203, opened 2016-08-11
FB messenger button

teehamaral:
This PR includes the Facebook Messenger button on the last position of the main menu (with internationalization) for non-global organizations.


norkans7:
The positioning of the button will not work for every org since some org have longer menu words, also some orgs might need to have Twitter instead or both.
We still wait for the design team to make a final decision about this that will work for all orgs

JPow:
Waiting on design team.  Trying to come up with a highly visible spot that doesn't distort whole tool bar.  Would say here the focus will be the Messenger button rather than Twitter - given some bad experience with Twitter bots and the coverage of Twitter for our target a lot of countries are moving away from it.

PR #204, opened 2016-08-19
adjustments to display new poll results faster for big accounts

norkans7:
This will calculate the results for poll and mark it as synced once we have pulled 100k runs and keep pulling until completed.

this will display the poll to the public sooner

### DIFF
--- a/ureport/backend/rapidpro.py
+++ b/ureport/backend/rapidpro.py
@@ -444,6 +444,9 @@ class RapidProBackend(BaseBackend):
                     fetch_start = time.time()
                     print "=" * 40
 
+                    if num_synced % 100000 == 0:
+                        poll.rebuild_poll_results_counts()
+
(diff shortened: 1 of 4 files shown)


norkans7:
@nicpottier please review

ureport/backend/rapidpro.py
```diff
@@ -444,6 +444,9 @@ def pull_results(self, poll, modified_after, modified_before, progress_callback=
                     fetch_start = time.time()
                     print "=" * 40
 
+                    if num_synced % 100000 == 0:
+                        poll.rebuild_poll_results_counts()
+
```
nicpottier: This isn't quite what I had in mind.. rather I was saying that after 100,000 polls we move onto the next poll that needs syncing. That way no single poll takes all the syncing for a period of time.

norkans7:
replaced by #206

PR #206, opened 2016-08-25
Poll sync improvements

norkans7:


ureport/backend/rapidpro.py
```diff
@@ -309,7 +309,7 @@ def pull_results(self, poll, modified_after, modified_before, progress_callback=
                 print "Start fetching runs for poll #%d on org #%d" % (poll.pk, org.pk)
```
nicpottier: For your own debugging / logging might want to include the cursor here..

ureport/backend/rapidpro.py
```diff
@@ -444,10 +444,24 @@ def pull_results(self, poll, modified_after, modified_before, progress_callback=
...
+
+                        print "Break pull results for poll #%d on org #%d in %ds, " \
+                              "created %d, updated %d, ignored %d" % (poll.pk, org.pk, time.time() - start, num_created,
+                                                                      num_updated, num_ignored)
+
+                        return num_created, num_updated, num_ignored
+
                 # update the time for this poll from which we fetch next time
                 cache.set(Poll.POLL_RESULTS_LAST_PULL_CACHE_KEY % (org.pk, poll.flow_uuid),
                           datetime_to_json_date(now.replace(tzinfo=pytz.utc)), None)
```
nicpottier: `after` really should be determined by the last record you've seen, not by now() since for all you know a new record was created between when you last fetched and now().

nicpottier:
looks good with those tweaks

ureport/backend/rapidpro.py
norkans7: we use now so that the next pull include results created while we were syncing

now is passed as before and the time saved in redis is used as after.Since the sync takes a while we save to redis the time when we started which was the latest at the time we started syncing

ureport/backend/rapidpro.py
norkans7: yep

ureport/backend/rapidpro.py
norkans7: may be I rename the variable from now to pull_started if you find this confusing

nicpottier: Understand why you are saving now, but I still don't think it is right, you may still be missing items. There is a gap between when you last fetched and now() by definition, and something could be created on rapidpro in between there.

That's why using the time of the last record fetched from rapidpro for that flow fixes that.

norkans7: I renamed the variable I think it name was confusing, the time we save is not timezone.now() when we finish but timezone.now() before we start pulling.

nicpottier: Oh, then I guess my question would be aren't you fetching more than you need in that case on the next sync?

norkans7: No since if a run is modified there are some new results to create.

The only case when it is unnecessary is for runs expirations which is why we are doing this
I guess we'll find a smarter way later how to ignore those

nicpottier: ok, go ahead and merge this but let's chat about this next week as I don't think you are getting what i"m saying.

norkans7: Sure,

PR #209, opened 2016-09-05
Link join steps

teehamaral:
This PR includes the option to add a link on the images of the steps on the page to join.


norkans7:
This does not seem to fit since those links are not on the site and it was not in the requirements.

I close this

PR #219, opened 2016-10-04
Update the existing map of objects when we update an a poll result

norkans7:


nicpottier:
This fixes the count issues?

norkans7:
yep, I think so

Issue #233, opened 2016-10-24
Error in location for participation chart


teehamaral:
Current:
![captura de tela 2016-10-24 as 11 10 37](https://cloud.githubusercontent.com/assets/5198870/19649388/01d7d278-99dc-11e6-9b7e-fdeed839e839.png)

How it should be:
![screen shot 2016-10-24 at 11 12 57](https://cloud.githubusercontent.com/assets/5198870/19649421/1f728260-99dc-11e6-9c45-a993ea2c63d7.png)

norkans7:
@teehamaral can you submit a PR with those fixes? Those seems to be broken because of the recent changes is styles and fonts.

teehamaral:
Yes @norkans7. I'll send the PR soon.

PR #234, opened 2016-10-24
Fix location participation chart

teehamaral:
This PR includes the fix to location for participation chart (#233)


norkans7:
Yep, thanks @teehamaral

PR #235, opened 2016-10-26
New translations

norkans7:


norkans7:
this might make us loose translations, should first update the PO file with those on Transifex

PR #241, opened 2016-11-01
update dash, and adjust tests

norkans7:


ureport/settings_common.py
```diff
@@ -419,14 +413,11 @@
...
 #-----------------------------------------------------------------------------------
 
-import djcelery
-djcelery.setup_loader()
-
 # by default, celery doesn't have any timeout on our redis connections, this fixes that
 BROKER_TRANSPORT_OPTIONS = {'socket_timeout': 5}
 
 
-CELERY_RESULT_BACKEND = 'djcelery.backends.cache:CacheBackend'
```
nicpottier: Why not set this to the REDIS_HOST as you do elsewhere?

PR #244, opened 2016-11-11
Django 1.9

norkans7:


pip-freeze.txt
```diff
@@ -8,15 +8,15 @@ contextlib2==0.5.4
 coverage==3.7.1
 rapidpro-dash==1.0.2
 dj-database-url==0.3.0
-Django==1.8.16
+Django==1.9.11
 django-appconf==1.0.1
 django-compressor==1.5
```
rowanseymour: Not 100% sure but I think compressor 2.0 is needed. Make sure your staging instance has compression enabled so you can test there

PR #245, opened 2016-11-15
Add lock timeouts

norkans7:


ureport/contacts/tasks.py
```diff
@@ -7,7 +7,7 @@
 logger = get_task_logger(__name__)
 
 
-@org_task('contact-pull')
+@org_task('contact-pull', 1800)
```
rowanseymour: This is only 30 mins - should be longer than longest possible time required

norkans7: I updated this to be 3 hours

Issue #247, opened 2016-11-16
Error on categories display when RTL language


teehamaral:
Current:
![screen shot 2016-11-16 at 15 09 01](https://cloud.githubusercontent.com/assets/5198870/20359536/1828fa6c-ac0f-11e6-843a-3704e6a5a371.png)

How it should be:
![screen shot 2016-11-16 at 15 11 51](https://cloud.githubusercontent.com/assets/5198870/20359538/1c776c8e-ac0f-11e6-8822-ad761f2d130c.png)

norkans7:
not clear to me the difference may be it'll be more clear with a PR

PR #248, opened 2016-11-16
Fix categories name on rtl language

teehamaral:
This PR includes the fixing referenced on #247


norkans7:
Ohh right now I get it, thanks @teehamaral

Issue #259, opened 2017-01-11
Some words aren't appearing in the tag cloud


teehamaral:
I have some words that aren't in the tag cloud, but they are in the database.
I confirmed this using the query below:
```
SELECT w.label, count(*) AS count 
FROM (SELECT regexp_split_to_table(LOWER(text), E'[^[:alnum:]_]') AS label 
FROM polls_pollresult WHERE polls_pollresult.org_id = my_org_id AND polls_pollresult.flow = 'my_flow_uuid' AND polls_pollresult.ruleset = 'my_ruleset_uuid' AND polls_pollresult.text IS NOT NULL AND polls_pollresult.text NOT ILIKE 'http%') w 
group by w.label
order by count DESC;
```
The "educación" word, for example, is returned about 72 times, but it isn't in the tag cloud.
In the code there a method to ignore some words, could it be the reason?

teehamaral:
Hi @norkans7. I understood your test. Already happened something like with this issue in your host? The word is in database but doesn't appear in tag cloud.

norkans7:
No missing words on our host. I am not sure if I get well the issue but to clarify on how the word cloud works; we take texts sent by reporters and break each into words and then filter the stop words(https://github.com/Alir3z4/python-stop-words)

norkans7:
Probably the reason it doesn't show is that it is not in the top 25 https://github.com/rapidpro/ureport/blob/master/templates/public/polls.haml#L390

teehamaral:
In `/pollquestion/{question_id}/results/` the word isn't show as well.

teehamaral:
@norkans7 I think that the `sorted` command (https://github.com/rapidpro/ureport/blob/master/ureport/polls/models.py#L770) should be before loop or add `order by count DESC` in SQL query. What do you think?

norkans7:
Right @teehamaral you can sort the ```uncleaned_categories``` before limiting to the first 100.
Please make a PR with the fix.

Does that solve your issue?

teehamaral:
Sure. Thanks @norkans7.

PR #260, opened 2017-01-12
test unicode word

norkans7:


norkans7:
@teehamaral I added a test for that and it seems to not be an issue like #259

PR #268, opened 2017-01-30
Fix words in tag cloud

### LINKED ISSUE #259: Some words aren't appearing in the tag cloud

teehamaral:
Order the words before limit to 100. Fix #259.


norkans7:
@teehamaral please add tests

norkans7:
Yep, thanks

PR #280, opened 2017-04-07
Update fr lang

teehamaral:
This PR includes the French language translation update (75%).


norkans7:
I pulled the latest translations from Transifex, so this is replaced by #285

Issue #288, opened 2017-05-09
Stories display 


nelsonlapf:
Hi guys,

Today the guys from Brazil asked us to hide the stories from the home page, however, it´s not possible unless if I remove them from the whole site, even from the story tab. 

When I uncheck the box "Is featured" in my opinion, the story should be "removed" from the home page and should continue to be displayed on the story tab, however, it´s the opposite when I unselect the box, the story is removed only from the story tab and still available on the home page. 

What do you think about this?

norkans7:
That makes sense as we changed to display the stories on the homepage the same way whether featured or not.

nelsonlapf:
that´s the point, I think that would better if we only display stories on the homepage only if it was set as featured and if it´s not, only be displayed on the stories tab.

norkans7:
Sure I agree we can remove the stories that are not featured to be shown on the homepage.
Please open a PR with those changes

PR #290, opened 2017-05-11
Stories display

teehamaral:
This PR is about the issue #288


norkans7:
@teehamaral @nelsonlapf I agree with the changes on the homepage but not with those on the stories page,

The stories page should remain as it is since the non featured stories are just listed under the section of 'Archived Stories'.

teehamaral:
Ok @norkans7, I'll revert the changes on the stories page

ureport/public/tests.py
```diff
@@ -335,26 +334,17 @@ def test_index(self, mock_cache_get):
 
         self.assertTrue(response.context['stories'])
         self.assertTrue(story1 in response.context['stories'])
-        self.assertFalse(response.context['other_stories'])
```
norkans7: I think you need to keep these tests as they were, actually this was testing stories from a different org are filtered out correctly, I would revert these tests and only adjust those that relate to 'other_stories' key

ureport/public/tests.py
```diff
@@ -455,6 +440,12 @@ def test_index(self, mock_cache_get):
         response = self.client.get(home_url, SERVER_NAME='nigeria.ureport.io')
         self.assertTrue('<div>INCLUDE MY CUSTOM HTML</div>' in response.content)
 
+    def test_additional_menu(self):
```
norkans7: +1 for more tests

norkans7:
Looks good now, thanks @teehamaral

PR #293, opened 2017-06-13
Result text field

norkans7:


ureport/public/tests.py
```diff
@@ -191,7 +191,7 @@ def test_has_better_domain_processors(self):
         # using subdomain wihout domain on org, login is shown and indexing should be allow
         response = self.client.get(home_url, HTTP_HOST='nigeria.ureport.io')
         self.assertEquals(response.request['PATH_INFO'], '/')
-        self.assertFalse('<meta content="noindex" name="robots">' in response.content)
```
rowanseymour: You should use `self.assertNotContains` and `self.assertContains` as these do proper decoding, and won't blow up when you get to python 3.

pip-freeze.txt
```diff
@@ -15,19 +14,21 @@ django-countries==3.3
 django-debug-toolbar==1.6
 django-digest==1.13
 django-guardian==1.4.2
+django-hamlpy==1.0
```
rowanseymour: Latest is 1.1.1

PR #294, opened 2017-06-15
Update deps

norkans7:


pip-requires.txt
```diff
@@ -1,5 +1,5 @@
 smartmin
-Django<1.9
+Django<2.0
 rapidpro-dash
 celery
 coverage
```
rowanseymour: you can remove `django-guardian` as well. If you still need the anon user then just copy the migration and util method from rapidpro

pip-requires.txt
```diff
@@ -24,7 +24,7 @@ django-redis
 psycopg2
 xlutils
 mock
-sorl-thumbnail==11.12.1b
+sorl-thumbnail==12.3
 hamlpy
```
rowanseymour: this should be replaced by django-hamlpy

pip-requires.txt
rowanseymour: Can also remove `django-nose` and `nose` if you're not using that for tests - we've not been using it elsewhere for a while now

ureport/assets/tests.py
```diff
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.core.urlresolvers import reverse
+from django.urls import reverse
```
rowanseymour: Is this a Django 2.0 change?

norkans7: django.core.urlresolvers module is deprecated in 1.10

rowanseymour: Ah didn't realise

PR #312, opened 2017-10-31
Limit the creation of polls to 1 per flow in 5 minutes

norkans7:


nicpottier:
Does Poll have an active field?

norkans7:
Yes

teehamaral:
Is there a reason for that? Just to know :)

norkans7:
@teehamaral we have notice some users on slow internet connection creating duplicated polls
and one without questions, this will prevent that

PR #316, opened 2017-11-10
Add flag to exclude fonts

norkans7:


nicpottier:
Would it make more sense to have a setting to let an org pick the font? this seems very specific without adding much flexibility

norkans7:
like a dropdown?

nicpottier:
I don't think we need a drop down, could just be the font declaration, basically inserting the styles. That would let them use whatever font they want and would override the default ones.

norkans7:
replaced by https://github.com/rapidpro/ureport/pull/317

PR #318, opened 2017-11-21
Makes dark section on home for all the picture

deimidis:
When height:auto is working, home page hero block isn't darker complete, but only the first part. This create a different behavior from other pages on the website.

Before
ureport_home_01

After
ureport_home_02


norkans7:
@deimidis I think this is so intentionally, we want the dark part only behind the poll title not all over the picture

PR #319, opened 2017-11-30
Support location path

norkans7:


norkans7:
Replaced by #320 before we move to better way of using using path field that will be added on RapidPro

PR #325, opened 2018-01-11
Reschedule poll results sync if the task failed with rate limit

norkans7:
Currently some orgs have many polls that `pull_results_other_polls` task is always failing because at some point a poll was hitting the rate limit. Here we will pause and reschedule the poll, and continue the other polls

### DIFF
--- a/ureport/polls/models.py
+++ b/ureport/polls/models.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 
 import logging
 import pytz
+import time
 from datetime import timedelta
 from django.contrib.auth.models import User
 from django.db import models, connection
@@ -13,6 +14,8 @@ from django.utils import timezone
 from smartmin.models import SmartModel
 from django.utils.translation import ugettext_lazy as _, ugettext
 from django.core.cache import cache
+from temba_client.exceptions import TembaRateExceededError
+
 from dash.orgs.models import Org
 from dash.categories.models import Category, CategoryImage
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


ureport/polls/models.py
```diff
@@ -147,14 +150,22 @@ def pull_results(cls, poll_id):
...
 
-        poll.rebuild_poll_results_counts()
+            Poll.objects.filter(org=poll.org_id, flow_uuid=poll.flow_uuid).update(has_synced=True)
 
-        Poll.objects.filter(org=poll.org_id, flow_uuid=poll.flow_uuid).update(has_synced=True)
+            return num_val_created, num_val_updated, num_val_ignored, num_path_created, num_path_updated, num_path_ignored
+        except TembaRateExceededError:
+            logger.error('Poll pull results exceeded rate error, rescheduling another pull results task',
+                         extra={'stack': True, })
+            time.sleep(120)
```
nicpottier: Why this sleep if we are doing a countdown below?

norkans7: I think we are in a for loop in the pull other polls task, so I wait a bit for the next poll

nicpottier: Why? Shouldn't we be updating as fast as we can or communicating back to the other task that this org has exceeded its quota so should try again later?

norkans7: I wanted to make the change in one place this is the method that is called by multiple tasks and I think changing those tasks is repetition(no DRY)

norkans7:
replaced by https://github.com/rapidpro/ureport/pull/326

PR #326, opened 2018-01-18
Use redis to record the last time a poll synced for long running 

norkans7:
Alternative approach for long running tasks like other_polls_task to skip poll that have a value set in redis that expires in 48 hours


ureport/settings_common.py
```diff
@@ -503,13 +503,13 @@
     },
     'results-pull-brick-polls':  {
         'task': 'dash.orgs.tasks.trigger_org_task',
-        "schedule": timedelta(hours=48),
+        "schedule": timedelta(hours=1),
```
norkans7: I increase the frequency the task run and we'll rely on the cache key flag that did not time out to skip the the pull until the key expire in 48 hours

PR #336, opened 2018-03-14
Add backend field on Contact, ContactField and Boundary

norkans7:


norkans7:
I think I can add the DB model so it store the backend API token(remove that from org config JSON)
and the backend config instead of adding them to the settings. 
that way an org could add a backend slug of an existing type without redeploying the code
and also store the host to be fetching the data

the settings would only have a list of backend types classes

@nicpottier does sounds easier to manage and worth?

nicpottier:
Ya, I think that makes sense, especially because I think we want to design this so that a single org may have two backends of type "floip". It isn't entirely unlikely that RapidPro might implement the floip standard for results one day, in which case we might have two floip backends for a single org, which would match that model a lot better than encoding them with a slug.

norkans7:
the way I think it will happen is that the slug will remain rapidpro will just switch the type,
currently it will be changing the settings file and push 

if we add the DB model then that will be via a migration or or the ui for each org.
And still no foreign keys o avoid joins.

I think there will be no time where RapidPro will connected using both floip and rapidpro types simultaneously. right?

nicpottier:
I'm not sure I understand, what are you proposing then?

norkans7:
The DB model will just store the configurations for the API  including the api token, the host to connect to and the type of backend  class to use and the backend slug like 'rapidpro', 'floip', 'providerX', 'provideY',...

the slugs will still be save on the Contacts, Poll, as it is now so we do not have joins in queries
And it will be easier to Add/modify the backend obj mostly switching the backend type.
We'll have views to manage those objects so no need to always push codes changes

nicpottier:
Hrmm.. I don't think that makes sense though. You could have two FLOIP backends no? The join isn't an expensive one since it is against a table of just a few rows so I'm not sure optimizing for that is saving much.

To me it seems like there should be a concept of a backend "type", which for now is just `rapidpro` and `floip`. And then orgs have one or more backends of particular types. So for now, they will have a `app.rapidpro.io` backend which is of type `rapidpro` and a `viamo` backend which is of type `floip`. But it is possible they could have both a `app.rapidpro.io` backend and a `textit.in` backend, both of type `rapidpro`.

Then each contact/flow/etc.. just has a foreign key to the backend it came from.

norkans7:
I think the names are confusing, what I call backend type is a class like https://github.com/rapidpro/ureport/blob/master/ureport/backend/rapidpro.py#L259

and the a backend is the source, 

so for Textit and RapiPro we can have 'textit' as RapidPro_backend_type and 'rapidpro' as backend the same type 

on the contacts we will always save `rapidpro` for data coming from `app.rapidpro.io` and `textit` for the data from `textit.in`

`viamo` will be of Type FLOIP and later `rapidpro` can be switched the type to use FLOIP class as well however the existing contacts will remain with field backend 'rapidpro' and should not change since I think changing the type should not change the data at all just the way we communicate to the service.

So even if rapidpro switch types the data should be similar.

nicpottier:
Alright well you seem to understand the issues there, so go for it and see how it turns  out!

PR #340, opened 2018-03-27
Add backend field

norkans7:
Replaces https://github.com/rapidpro/ureport/pull/336


nicpottier:
I'm ok with this, though I don't particularly agree with the approach, but I think it'll be a good lesson one way or another for you. (I would rather have a foreign key to backend instead of a backend slug, but you should get to try to do it your way)

PR #343, opened 2018-03-27
Do not add backend field on poll results

norkans7:


ureport/polls/migrations/0051_poll_backend.py
```diff
@@ -8,12 +8,12 @@
...
+        ('polls', '0050_auto_20170615_1455'),
     ]
 
     operations = [
-        migrations.AlterField(
-            model_name='pollresult',
+        migrations.AddField(
+            model_name='poll',
             name='backend',
             field=models.CharField(default='rapidpro', max_length=16),
```
nicpottier: I think you want no default here.

PR #346, opened 2018-04-04
Switch to use rapidpro config

norkans7:
* Switch to use config from 'rapidpro' key in config
* Remove unused and unnecessary codes


nicpottier:
What is `top_key` in these? The path to the config? Would it be cleaner to allow paths in the keys? IE `rapidpro.reporter_group`?

norkans7:
top_key is the top level key of the org.config dict, so far we have "common" and "rapidpro"

how to allow those paths? by splitting the key by `.` ?

PR #350, opened 2018-04-05
Backend foreign key

norkans7:


ureport/contacts/migrations/0017_auto_20180405_1330.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contacts', '0016_auto_20180314_1317'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
norkans7: I just remove the existing fields and added one
I did not keep the data since I think there is not much value to migrate those

ureport/contacts/migrations/0018_auto_20180405_1336.py
```diff
@@ -0,0 +1,54 @@
...
+
+    start = time.time()
+
+    for org in Org.objects.all():
+        backend = org.backends.filter(slug="rapidpro").first()
+        ContactField.objects.filter(org=org).update(backend=backend)
+        print("Processed fields")
+
+        i = 0
+        contact_ids = Contact.objects.filter(org_id=org.id).values_list('id', flat=True)
```
norkans7: Process 1000/sec so it will take around 90min for 5M contacts

norkans7: I think we can pause syncing and migrate these in a screen

PR #352, opened 2018-04-09
Make sure tasks do no sync polls synced by other tasks that run more …

norkans7:
…often

### DIFF
--- a/ureport/polls/models.py
+++ b/ureport/polls/models.py
@@ -358,8 +358,10 @@ class Poll(SmartModel):
     def get_recent_polls(cls, org):
         now = timezone.now()
         recent_window = now - timedelta(days=7)
+        main_poll = Poll.get_main_poll(org)
 
-        recent_other_polls = Poll.get_public_polls(org).exclude(created_on__lte=recent_window).order_by('-created_on')
+        recent_other_polls = Poll.get_public_polls(org).exclude(pk=main_poll.pk)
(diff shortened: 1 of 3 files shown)


ureport/polls/tasks.py
```diff
@@ -89,8 +92,12 @@ def pull_results_brick_polls(org, since, until):
 def pull_results_other_polls(org, since, until):
     from .models import Poll
 
+    now = timezone.now()
+    recent_window = now - timedelta(days=7)
+
     results_log = dict()
-    other_polls_ids = Poll.get_other_polls(org).order_by('flow_uuid').distinct('flow_uuid').values_list('id', flat=True)
+    other_polls_ids = Poll.get_other_polls(org).exclude(created_on__gt=recent_window)
```
norkans7: recent polls are synced more often that these here so exclude them

ureport/polls/models.py
```diff
@@ -358,8 +358,10 @@ def get_other_polls(cls, org):
     def get_recent_polls(cls, org):
         now = timezone.now()
         recent_window = now - timedelta(days=7)
+        main_poll = Poll.get_main_poll(org)
 
-        recent_other_polls = Poll.get_public_polls(org).exclude(created_on__lte=recent_window).order_by('-created_on')
+        recent_other_polls = Poll.get_public_polls(org).exclude(pk=main_poll.pk)
```
norkans7: this method is used by the task only so it is safe to change here

the main poll syncs more often by the main_poll_task, we should not sync it again here.

PR #353, opened 2018-04-25
set a default past date time for registered_on so PG triggers work co…

norkans7:
…rrectly

### DIFF
--- a/ureport/backend/rapidpro.py
+++ b/ureport/backend/rapidpro.py
@@ -202,7 +202,7 @@ class ContactSyncer(BaseSyncer):
                 registered_on = json_date_to_datetime(registered_on)
 
         if not registered_on:
-            # default to creted_on to avoid null in the PG triggers
+            # default to created_on to avoid null in the PG triggers
             registered_on = json_date_to_datetime(remote.created_on)
 
         occupation = ''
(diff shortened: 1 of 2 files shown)


nicpottier:
Wouldn't setting that to now() be more accurate than a random date in the past? Seems like setting to 2001 will lead to some weird data no?

norkans7:
Our graph shows up to six month in the past I thought using that old date is just fine for this case
using now will assume the contact was just registered so all contact with no date set will appear as if they registered today, Isn't that more confusing?

nicpottier:
We have a created_on on the contact no? Maybe use that? That seems likely to be much closer than either.

norkans7:
Okay

ureport/backend/rapidpro.py
```diff
@@ -202,8 +202,8 @@ def local_kwargs(self, org, remote):
                 registered_on = json_date_to_datetime(registered_on)
 
         if not registered_on:
-            # default to a date in in the past to avoid null in the PG triggers
-            registered_on = json_date_to_datetime('2000-01-01T00:00:00.000')
+            # default to creted_on to avoid null in the PG triggers
```
nicpottier: typo

PR #354, opened 2018-04-25
Gin index

norkans7:
This will address https://sentry.io/nyaruka/ureport/issues/532862273 

I'll do more experiments to make sure this does not degrade the current performance


norkans7:
@nicpottier experimented with the index locally the poll result will be saved correctly

norkans7:
the other option to solve this is to keep the existing index and just limit the text to shorter that 2560 for example

nicpottier:
Oh these are for word clouds huh? Ya, we should just trim those to something more reasonable as a first go.

norkans7:
Will do the trim on another branch

PR #358, opened 2018-05-16
Custom Welcome Text on Facebook Plugin

johncordeiro:
This PR adds a custom welcome text to the Facebook Messenger plugin.

PR #360, opened 2018-05-23
Add configuration for Facebook Messenger Customer Chat plugin message

norkans7:


norkans7:
Already addded in #358

PR #363, opened 2018-05-28
Change text editor library  and use summernote

norkans7:
Requires https://github.com/rapidpro/dash/pull/106


norkans7:
text-editor

norkans7:
The old one was broken

PR #366, opened 2018-06-06
Hotfix organization language

johncordeiro:
This PR fixes the org language context variable, by setting up the right iso code depending on org language. Today is only working for RTL languages.


ureport/public/context_processors.py
```diff
@@ -75,8 +75,9 @@ def set_org_lang_params(request):
     is_rtl_org = False
     org_lang = 'en_US'
     org = request.org
-    if org and org.language in getattr(settings, 'RTL_LANGUAGES', []):
-        is_rtl_org = True
+    if org and org.language:
+        if org.language in getattr(settings, 'RTL_LANGUAGES', []):
```
norkans7: why not one if block?

johncordeiro: hey @norkans7, in this case, the org_lang will be updated only for RTL_LANGUAGES otherwise all the other languages keeps the default en_US. That's why countries that are configured as Spanish still shows up "Message us" in the Facebook Button, for example.

norkans7: Alright, I get it now. Please add a test for that
may be here https://github.com/rapidpro/ureport/blob/0318f754742268ea0f3e2a92f0ac04c5843d84fc/ureport/public/tests.py#L215

norkans7:
Looks good, thanks

PR #368, opened 2018-06-14
Sort pie charts from the most recent to oldest ones

johncordeiro:
This PR adds the order by `created_on` when querying recent polls.


johncordeiro:
any thoughts on that @norkans7 ?

PR #372, opened 2018-07-19
Use archives

norkans7:


nicpottier:
Nice, this looks nice and simple. Have you tried it on staging at all yet to see what performance is like?

norkans7:
I was able to pull one poll in a bit more than an hour  around 4000sec

nicpottier:
What poll was that? Are the results the same as production?

PR #383, opened 2018-10-10
Add results grouped by gender and by age to the API

norkans7:


ureport/api/tests.py
```diff
@@ -373,6 +373,8 @@ def test_single_poll(self):
                             ruleset_uuid="uuid1",
                             title="What's on mind? :)",
                             results=dict(set=20, unset=10, open_ended=False, categories="CATEGORIES-DICT"),
+                            results_by_age=[dict(set=20, unset=10, open_ended=False, categories="CATEGORIES-DICT")],
```
nicpottier: What do the values here represent? This is an array, but what are the keys into the array? Maybe this should be a dict of dicts?

norkans7: Those are from the mock

PR #402, opened 2019-03-07
Fix date format in News

johncordeiro:
This PR fixes the date format in News listing.


johncordeiro:
@norkans7 any thoughts on that?

norkans7:
So this will localize using the browser language, right?

PR #404, opened 2019-03-13
🚒 Simultaneously support both old engine and new engine flow definitions

norkans7:

### DIFF
--- a/ureport/backend/rapidpro.py
+++ b/ureport/backend/rapidpro.py
@@ -7,6 +7,7 @@ import json
 import logging
 import time
 from collections import defaultdict
+from distutils.version import StrictVersion
 
 import pytz
 import requests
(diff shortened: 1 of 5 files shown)


ureport/backend/rapidpro.py
```diff
@@ -329,47 +330,77 @@ def get_definition(self, org, flow_uuid):
         export_definition = client.get_definitions(flows=(flow_uuid,))
```
rowanseymour: if you include `dependencies=none` won't this always return a single flow and thus you don't need the loop below?

ureport/backend/rapidpro.py
```diff
@@ -329,47 +330,77 @@ def get_definition(self, org, flow_uuid):
...
-                is_active=False
-            )
+            flow_nodes = flow_definition["nodes"]
+
+            for node in flow_nodes:
+                if "router" not in node:
+                    continue
+
+                question = PollQuestion.update_or_create(
+                    user, poll, node["router"].get("result_name", ""), node["uuid"], node["router"]["type"]
```
rowanseymour: the types here (switch vs random) aren't equivalent to types on legacy rulesets (wait_response, etc)

maybe you need to look at the `wait` on the router? `wait{type=msg}` is the equivalent of a router of type wait_response

norkans7: I think any node with a router can be added to U-Report so anything that was a ruleset even sllit by contact field

rowanseymour: I'm not sure what the rules are for what equals a poll question - just giving a heads up that `router.type` isn't the same thing as `ruleset.ruleset_type`. The former determines how the router picks a path. Then the `wait` on the node is what determines what we're waiting for, i.e. a response

nicpottier: I can see arguments either way but ya I would generally agree with Rowan that we should only be showing those splits that are waits as well. Do we have any U-Report polls that use a split by expression or the like as a poll response that they graph and display?

We will probably revisit how we display these things in the upcoming v2 of U-Report, so being inclusive for the time being may be just fine. (IE any split that saves a result)

norkans7: I used exits to create the categories and I considered any node with router to be a poll question,
I think that will allow displaying results by contacts fields like not necessarily wait for msg, and admin can remove the questions they want to not publish

PR #405, opened 2019-04-03
Add /v2 urls for public site, site under construction page

norkans7:


norkans7:
U-Report 2019-04-03 19-11-34

All v2 pages are currently like ⬆️

norkans7:
Yep, we will still need to have the more custom styles depending on org as well, I think we can use a different stylesheet as we currently do

PR #406, opened 2019-05-02
Update django.po

JesusEkie:
An improved translation to better reflect the original message.


norkans7:
@JesusEkie The best way is to add those to https://www.transifex.com/rapidpro/ureport/

PR #418, opened 2019-07-23
U-Reporters page graphics issue when laguages uses ' for region's names.

WilliamTales:
# Description

At U-Reporters page for countries which has ' within regions' name, the page script bugs and graphics don't load up.
EG. **Uzbekistan** which in **Uzbek** has **Farg'ona viloyati** as a region name.
The page's graphics stays loading https://uzbekistan.ureport.in/ureporters 

# Proposed solution

Replacing codes ' by ".


WilliamTales:
Hey @nicpottier your attention is highly appreciated.

PR #428, opened 2019-09-05
CSS tweaks

nicpottier:
@norkans7 started down a rabbit hole of refactoring the css a bit. I think it is getting there but didn't make it much farther than first half of index page. Going to continue tomorrow but this seems to be working well from a responsive perspective and I think styles are getting simpler.

Still a bit of margin on the right that I'm fighting with UI kit over. Wondering whether we can remove their JS stuff because I think that is causing us a lot of grief to some extent.. will investigate.

Just out of curiosity, what do you feel UI kit is buying us? Just the grid stuff? I feel like maybe we'd just be better off with straight flexbox.


nicpottier:
Also don't merge this yet, want to get it farther.

norkans7:
@nicpottier the reason I had started to use UIKit was because I kept getting requests to use https://getuikit.com/docs/scrollspy, the accordion behavior(even tho I kept the custom one we could switch to use the built in for UIKit) and the other animations so I thought the design was inspired with it or made considering it, It made sense for me to switch to it

PR #435, opened 2019-09-08
Add engagement graphs on the engagement page with tailwind CSS designs

norkans7:


templates/v2/public/polls.haml
```diff
@@ -41,21 +41,20 @@
...
 
-            {% if org|config:"has_extra_gender" %}
-            .text-center.inline-block.pl-4.border-l.border-black
-              %img(src="{{ STATIC_URL }}img/gender-other.png" width="40")
-              .font-bold.text-xl.leading-none.mt-2
-                {{gender_stats.other.percentage|default:"0"}}%
-              .text-sm.leading-none
-                {{gender_stats.other.count|default:"0"|intcomma}}
-            {% endif %}
+            - if org|config:"has_extra_gender"
```
nicpottier: No need to change this right now but I've found it is a bit clearer to not use indentation for `if`/`for` in django templates in haml just because then the structure of the tags is a bit clearer. (everything at the same indentation is the same level in the HTML tree instead of them being more indented just for the if/fow but really being at the level of their parent)

templates/v2/public/ureporters.haml
```diff
@@ -241,6 +259,321 @@
       });
     });
 
+    $(function(){
```
nicpottier: I started pulling all this stuff out into their own coffee files and not having much of any JS in the page. Main reason is that coffee will compile to something compatible across browsers which it is still easy to leave a trailing comma that breaks IE etc.. 

I'll see if I can generalize out the pill selection behavior.

PR #444, opened 2019-09-10
Admin nav public site

norkans7:


templates/public_base_v2.haml
```diff
@@ -125,6 +125,132 @@
 
     // top black bar
     .block.bg-black.h-2
+    
+    // admin nav
+    - if request.user.is_authenticated and request.org in request.user.get_user_orgs or request.user.is_superuser
+      .block.hidden(class="md:block")
+        %ul.flex.bg-gray-800.text-white
+          %li
+            %a.inline-block.px-6.py-4(href="{% url 'orgs.org_home' %}" class="hover:text-gray-800 hover:bg-gray-300")
```
nicpottier: When you have a set of styles repeated this many times you are probably best using a class with `@apply`

PR #447, opened 2019-09-10
remove unused codes

norkans7:


norkans7:
Right, I'll move the script to coffee files in a different PR

PR #449, opened 2019-09-11
add state segmenting to polls page

nicpottier:


templates/v2/public/polls.haml
```diff
@@ -132,10 +132,22 @@
...
+              .state-dropdown.absolute.z-50(id="states-{{question.id}}" style="margin-left:-1px")
+                .border.border-gray-400.text-left.bg-white.h-64(style="overflow-y:scroll;margin-top:.5rem;margin-left:-1rem;")
+                  .pt-1.text-black
+                    {% for state in states %}
+                    .pl-2.pr-4.py-1.state-segment(data-state='{{state.id}}' data-question="{{question.id}}" class="hover:bg-gray-200 {% if forloop.counter0 < 3 %}selected{% endif %}")
+                      .checkbox.float-left
+                        %i.fa.fa-check
+                      .ml-4.whitespace-no-wrap
+                        {{ state.name }}
+                    {% endfor %}
```
norkans7: I think we need to keep using the haml tags way 

```
-for state in states
```
and omit the closing tags

nicpottier: Why do we need to?

templates/v2/public/polls.haml
```diff
@@ -140,13 +140,12 @@
...
-                        {{ state.name }}
-                    {% endfor %}
+                    - for state in states
+                      .pl-2.pr-4.py-1.state-segment(data-state='{{state.id}}' data-question="{{question.id}}" class="hover:bg-gray-200 {% if forloop.counter0 < 3 %}selected{% endif %}")
+                        .checkbox.float-left
+                          %i.fa.fa-check
+                        .ml-4.whitespace-no-wrap
+                          {{ state.name }}
 
           {% else %}
```
norkans7: Same for `if/else` ?

nicpottier: For now lets merge, I'll change everything based on https://twist.com/a/89176/ch/222296/t/1052756/

PR #457, opened 2019-09-13
Reduce font-size fo r longer texts

norkans7:


templates/v2/public/polls.haml
```diff
@@ -49,7 +49,7 @@
           .bg-dark2.inline-block.px-4.py-2.uppercase.leading-none.font-bold.text-sm.text-white
             {{ latest_poll.category.name }}
 
-          .text-5xl.font-bold.leading-tight
+          .font-bold.leading-tight(class="{% if latest_poll.title|length < 50 %}text-5xl{% else %}text-4xl{% endif %}")
```
nicpottier: We should probably be consistent here with homepage, let's do 4xl and 3xl then.

PR #459, opened 2019-09-16
No location config

norkans7:


templates/v2/public/ureporters.haml
```diff
@@ -63,12 +63,13 @@
...
-          .text-sm.font-bold.mb-2
-            -trans "U-REPORTERS BY REGION"
-          #ureporter-map.h-full.w-full.map.block(data-map-url="/v2/contact_field_results/" data-district-zoom="'{{ district_zoom|lower }}' === 'true'" data-ward-zoom="'{{ ward_zoom|lower }}' === 'true'" style="min-height:300px;min-width:300px;")
-          #map-info.relative.h-full.w-full.mt-2.flex.justify-center
+      -if show_maps
+        .flex-1.text-white.bg-dark-blue.vertical-angle
+          .pl-6.pr-home.py-home.text-center
+            .text-sm.font-bold.mb-2
+              -trans "U-REPORTERS BY REGION"
+            #ureporter-map.h-full.w-full.map.block(data-map-url="/v2/contact_field_results/" data-district-zoom="'{{ district_zoom|lower }}' === 'true'" data-ward-zoom="'{{ ward_zoom|lower }}' === 'true'" style="min-height:300px;min-width:300px;")
```
nicpottier: I don't think this is valid HTML/JS. Data attributes aren't generally executed javascript. Maybe it works in Chrome but is this a web standard?

norkans7: I am not sure

norkans7: I change that to use django

norkans7:
Included in #461

PR #486, opened 2019-10-04
Added "sitestatic" in middleware

edudouglas:


norkans7:
What is this fixing?

edudouglas:
To prevent redirection of all statistical files, for example "/sitestatic/*.css" to "/v2/sitestatic/*.css".

norkans7:
Can you show were that happened on the live server?
I do not think that is needed for the sitestatic files

edudouglas:
In our case the U-Report project runs in docker container and Nginx works as a reverse proxy outside the container, redirecting any HTTP access to the container, so it is not possible to map Nginx to the "sitestatic" entry as it is done in this line https://github.com/rapidpro/ureport/blob/master/nginx.conf#L33, so the solution was to add "sitestatic" in the middleware, so it will not redirect static files to "/v2/sitestatic/".

norkans7:
May be I am missing something but I think we are not calling the middleware for those static files, but again the middleware is a temporal solution before we switch all sites to the new version so I think that is fine to have sitestatic skipped too

PR #495, opened 2019-10-07
Add translations for key words in spanish

johncordeiro:


norkans7:
@johncordeiro the issue with this approach is that it will be in conflicts with Transifex
Can you make those changes on https://www.transifex.com/rapidpro/ureport/
and Transifex will push those in a PR like https://github.com/rapidpro/ureport/pull/492

johncordeiro:
Hey @norkans7, it was already done in Transifex. Do you know why no PR was received yet?

norkans7:
then some strings are untranslated, the PR is open when we get 100% translated strings

Issue #523, opened 2019-11-08
Single widget pages


WilliamTales:
https://github.com/rapidpro/ureport/blob/a2392637a8d2f02f2165251819166682570528db/templates/v2/public/index.haml#L258

When a country uses only one widget its **div** isn't being displayed because it's missing **open** parameter.

PR #524, opened 2019-11-08
fixed parameter when only one social media widget is used

WilliamTales:
#523 
This fixes the parameter when only one social media widget is used.

### DIFF
--- a/templates/v2/public/index.haml
+++ b/templates/v2/public/index.haml
@@ -254,8 +254,8 @@
   :javascript
     $(function(){
     if($(".social-media-blocks").children().length == 1) {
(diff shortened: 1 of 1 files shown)


templates/v2/public/index.haml
```diff
@@ -253,6 +253,10 @@
 
   :javascript
     $(function(){
+    if($(".social-media-blocks").children().length == 1) {
+        $(".social-media-blocks").children().toggleClass("open", 1000, "easeOutSine");
```
norkans7: I think we want that to be `AddClass` instead of `toggleClass` ?

PR #526, opened 2019-11-19
Add limit poll states config for superusers

johncordeiro:
We've faced a very specific use case in Serbia, where some states need to appear on the map, but they couldn't appear in the poll results (for political reasons). This PR keeps the limit states config to remove states in the engagement map but creates a new config for limiting states in poll results.

### DIFF
--- a/ureport/settings_common.py
+++ b/ureport/settings_common.py
@@ -392,6 +392,11 @@ ORG_CONFIG_FIELDS = [
         superuser_only=True,
     ),
     dict(
+        name="limit_poll_states",
+        field=dict(help_text=_("The states to show on maps only, used to filter poll results"), required=False),
+        superuser_only=True,
+    ),
+    dict(
         name="google_tracking_id",
         field=dict(
             help_text=_("The Google Analytics Tracking ID for this organization"),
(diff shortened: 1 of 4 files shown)


ureport/settings_common.py
```diff
@@ -391,6 +391,11 @@
         field=dict(help_text=_("The states to show on maps only"), required=False),
         superuser_only=True,
     ),
+    dict(
```
norkans7: I think we can reuse the existing "limit_states" config and make sure we filter them properly everywhere is needed

johncordeiro: Yes, but the use case here is very specific. Let's think that some countries have conflict areas on the map, and when joining the website they see the Country map. The office wants to show the complete map, even though some areas that do not collect data for U-Report, for political reasons.

However, when you go to poll results in these areas, they do not have any data collected, but they are in results by location, and also they appear already checked in the location list.

That's why I've decided to create a new config, to show some areas in the map (as they are part of the country), but to not show any poll results for them. Initially, I thought exactly like you, but it didn't fill the use case.

norkans7: @johncordeiro, I think a location not having data is fine and does not need to be hidden

PR #527, opened 2019-11-19
Eenable locale for Serbian Latin Serbia language

matmsa27:
Hi everyone, I open this PR to have available the Serbian Latin Language (`sr_RS@latin`) on the UReport. This locale on transifex has 100% translated.

### DIFF
--- a/ureport/settings_common.py
+++ b/ureport/settings_common.py
@@ -82,7 +82,6 @@ LANGUAGES = (
     ("it", "Italian"),
     ("ro", "Romanian"),
     ("vi", "Vietnamese"),
-    ("sr-latn", "Latin Serbian"),
     ("sr-rs@latin", "Serbian (Latin, Serbia)"),
 )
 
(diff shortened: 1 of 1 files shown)


ureport/settings_common.py
```diff
@@ -83,6 +83,7 @@
     ("ro", "Romanian"),
     ("vi", "Vietnamese"),
     ("sr-latn", "Latin Serbian"),
+    ("sr-rs@latin", "Serbian (Latin, Serbia)"),
```
norkans7: What is the previous one("sr-latn") for?

norkans7: I think we only need one of them not both

matmsa27: There are the two folders on locale for Serbia Language, the sr-rs@latin has 100% translated on transifex. I agree it's necessary one of them and is good for UReport use the locale with more translations (`sr-rs@latin`). We can disable the sr-latn and remove locale(`sr-latn`) option for the languages list, or if necessary remove the folder too.

norkans7: @matmsa27 right you just keep one Serbian option and its folder of locale files

matmsa27: Right @norkans7 , it's done!

PR #533, opened 2019-11-22
Fixing on image responsivity while scaling large images on story detail

periclesJ:
Large images on small devices becomes stretched, this small fix solves this problem keeping image aspect ratio inside 'story detail' on v1.

### DIFF
--- a/templates/public/story_read.haml
+++ b/templates/public/story_read.haml
@@ -156,7 +156,9 @@
     }
 
     .story-details img {
-      max-width: 100%;
+      height: auto !important;
+      display: block;
+      width: 100% !important;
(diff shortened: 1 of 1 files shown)


norkans7:
Any screenshot to see the issue vs the fix?

pericles-jr:
This the layout before fixing:
Captura de Tela 2019-11-25 às 08 31 00

This is the layout after the adjustment:
Captura de Tela 2019-11-25 às 08 32 03

norkans7:
Thanks, I see that issue.
FWIW it is recommended you switch to the new design version since the old one will be removed soon

PR #542, opened 2019-12-11
Add method to squash poll stats

norkans7:


norkans7:
Replaced by #550 and #551

PR #545, opened 2019-12-23
Remove numbers on string translated

matmsa27:


locale/ro/LC_MESSAGES/django.po
```diff
@@ -1378,7 +1378,7 @@ msgstr "TOATE RĂSPUNSURILE "
 
 #, python-format
 msgid "%(responded)s responded out of %(polled)s polled"
-msgstr " 1 %(responded)s răspondenți din 2 %(polled)s intervievați"
+msgstr "%(responded)s răspondenți din %(polled)s intervievați"
```
norkans7: This should be done on Transifex instead

matmsa27: Yeah, when I open this PR, the transifex for `ro` has 100% translated, I tried to change the string on transifex and the PR wasn't created.

That string translated already done on transifex, but the PR wasn't created.
Screenshot from 2020-01-10 10-41-15

Actually the transifex for `ro` presents 94% translated, but when try to translate doesn't show the strings for translate.
Screenshot from 2020-01-10 10-50-47

PR #548, opened 2020-01-14
Squash stats with field

norkans7:


ureport/stats/migrations/0007_pollstats_is_squashed.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('stats', '0006_populate_word_clouds'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='pollstats',
+            name='is_squashed',
+            field=models.BooleanField(default=False, help_text='Whether this row was created by squashing'),
```
nicpottier: If there are a ton of rows here, then populating this is going to take a very long time since it isn't null. Better to make it nullable and treat null and False the same way.

norkans7:
Replaced by #550 and #551

PR #580, opened 2020-04-30
Fix size from bulgaria flag

matmsa27:
I open this pr to fix the size from Bulgaria flag.

Currently, it is like this:

Clipboard - 29 de Abril de 2020 às 18_11

With the fix, will be like this:
Screenshot from 2020-04-29 21-53-21


norkans7:
Thanks @matmsa27

PR #593, opened 2020-06-02
Create users with the same UUID in different orgs

hudsonbrendon:
We are working on a project with Venezuelan refugees in three different countries in Latin America. The aim of the project is to provide support and listen to refugees who are migrating between these countries. There is only one workspace in rapidpro connected to 3 dashboards in the U-report. The workspace is managed by the three countries, but each has an environment in the u-report to drop polls separately. The registered number must be the same in the different dashboards. Because of this rule, the data synchronization of u-reports was not occurring in the correct way, causing discrepancy in the data in the different dashboards. The system does not allow the creation of users with the same UUID in different orgs, this is making it impossible to register and manage these refugees. After checking the code, we identified the problem and this is our correction proposal.


norkans7:
the changes looks good make the tests pass and I guess the issue is the imports moved

hudsonbrendon:
@norkans7 I solved the conflicts, is there any more addition or correction in the PR to be made?

ureport/contacts/tests.py
```diff
@@ -25,6 +25,17 @@ def setUp(self):
         self.nigeria.set_config("rapidpro.female_label", "Female")
         self.nigeria.set_config("rapidpro.male_label", "Male")
 
+        self.uganda.set_config("rapidpro.reporter_group", "Ureporters")
```
norkans7, line 28: I do not think these changes here are needed

ureport/contacts/tests.py
```diff
@@ -73,6 +84,53 @@ def setUp(self):
         self.born_field = ContactField.objects.create(org=self.nigeria, key="born", label="Born", value_type="T")
         self.gender_field = ContactField.objects.create(org=self.nigeria, key="gender", label="Gender", value_type="T")
 
+        self.country = Boundary.objects.create(
```
norkans7, line 87: Same with these, I think this is not testing anything, remove them back

ureport/contacts/tests.py
```diff
@@ -89,6 +147,26 @@ def test_get_or_create(self):
         self.assertEqual(created_contact.pk, existing_contact.pk)
         self.assertEqual(existing_contact.born, 2000)
 
+    def test_create_with_same_uuid(self):
```
norkans7, line 150: On this test I think we need to test the ID are different even if the UUID are the same 
Also Add a test that shows we can not have two objects on the same org with the same UUID

ureport/contacts/tests.py
```diff
@@ -89,6 +147,26 @@ def test_get_or_create(self):
...
+        self.assertIsNone(Contact.objects.filter(org=self.nigeria, uuid="contact-uuid").first())
+        self.assertIsNone(Contact.objects.filter(org=self.uganda, uuid="contact-uuid").first())
+
+        created_contact_nigeria = Contact.get_or_create(self.nigeria, "contact-uuid")
+
+        created_contact_uganda = Contact.get_or_create(self.uganda, "contact-uuid")
+
+        self.assertEqual(created_contact_uganda.uuid, created_contact_nigeria.uuid)
+
+    def test_create_with_different_uuid(self):
```
norkans7, line 160: this is not a case to worry about really I think but. you can keep that in the same test method any way instead of a different method those are all related

hudsonbrendon:
@norkans7 Any more requests to be made?

ureport/contacts/tests.py
```diff
@@ -89,6 +99,33 @@ def test_get_or_create(self):
...
+        self.assertEqual(created_contact_uganda.uuid, created_contact_nigeria.uuid)
+
+        Contact.objects.all().delete()
+
+        created_contact_nigeria = Contact.get_or_create(self.nigeria, "contact-uuid-nigeria")
+        created_contact_uganda = Contact.get_or_create(self.uganda, "contact-uuid-uganda")
+
+        self.assertNotEqual(created_contact_nigeria.uuid, created_contact_uganda.uuid)
+
+        Contact.objects.all().delete()
```
norkans7, line 121: That is a duplicate of the test above I think these can also be removed

ureport/contacts/tests.py
```diff
@@ -25,6 +25,16 @@ def setUp(self):
...
+        self.uganda.set_config("rapidpro.registration_label", "Registration Date")
+        self.uganda.set_config("rapidpro.state_label", "State")
+        self.uganda.set_config("rapidpro.district_label", "LGA")
+        self.uganda.set_config("rapidpro.ward_label", "Ward")
+        self.uganda.set_config("rapidpro.occupation_label", "Activité")
+        self.uganda.set_config("rapidpro.born_label", "Born")
+        self.uganda.set_config("rapidpro.gender_label", "Gender")
+        self.uganda.set_config("rapidpro.female_label", "Female")
+        self.uganda.set_config("rapidpro.male_label", "Male")
+
```
norkans7, line 37: Why are these added? I think we do not need these either

hudsonbrendon:
@norkans7 Any more requests to be made?

PR #607, opened 2020-07-20
Enable croatian language

matmsa27:


ureport/settings_common.py
```diff
@@ -84,6 +84,7 @@
     ("vi", "Vietnamese"),
     ("sr-rs@latin", "Serbian (Latin, Serbia)"),
     ("bg", "Bulgarian"),
+    ("hr", "Croatian"),
```
norkans7, line 87: @matmsa27 should this be `hr_HR` as what is submitted to Transifex ?

matmsa27: Hi @norkans7 , ya I fixed this on the last commit

matmsa27:
Thanks @norkans7

PR #615, opened 2020-07-31
Add sync schedule on opinions admin list

norkans7:


norkans7:
U-Report 2020-07-31 14-37-46

I changed the table to have room for one more column

norkans7:
I think that is the first step and not allowing to manually resync yet

norkans7:
BEFORE opinions table
U-Report 2020-07-31 15-39-28

nicpottier:
Looks like a reasonable first step, should probably use plural translation to get rid of awkward `is/are`

PR #622, opened 2020-09-03
[FIX]: temba_client version to v2 on migration files

paulobernardoaf:
- The temba_client version was outdated, and the `python manage.py migrate` command was failing because the correct version is temba_client.v2.


norkans7 (approved):

norkans7:
Looks good

PR #634, opened 2020-09-29
Fix: code_check for translations_django-po--master_sr_RS@latin

paulobernardoaf:
Hey @norkans7 , this PR is to fix the #632 made from Transifex that did not run the code_check. As we don't have the permission to push directly into that branch.


norkans7:
Thanks @paulobernardoaf

PR #642, opened 2020-11-03
Shared flags

norkans7:


ureport/utils/__init__.py
```diff
@@ -211,16 +204,61 @@ def update_poll_flow_data(org):
                         Poll.objects.filter(pk=poll.pk).update(**updated_fields)
 
 
+def fetch_shared_sites_count():
+    import requests
+
+    this_time = datetime.now()
+    try:
+        response = requests.get("https://ureport.in/shared_sites_count/")
```
norkans7, line 212: this is the place all installs will use the common settings from our install.

PR #661, opened 2020-12-09
Fix stories list in index page

Sandro-Meireles:
When creating a new story in `manage/story/` it should appear on the home page, as the first item on the list, however, this does not happen. On the home page, stories are being listed incorrectly, starting from the second item.

- The problem is:
Captura de Tela_Área de Seleção_20201209092526
Captura de Tela_Área de Seleção_20201209093227

To solve this problem I changed the filter responsible for listing them.

- in `templates/public_base.haml` line `537`
```
-        - for story in main_stories|slice:"1:4"
+        - for story in main_stories|slice:"0:3"
```

### DIFF
--- a/templates/public_base.haml
+++ b/templates/public_base.haml
@@ -530,11 +530,11 @@
      
       -block content
   
-      {% if main_stories|slice:"1:" %}
+      {% if main_stories|slice:"0:" %}
       // footer stories
       .bg-dark-blue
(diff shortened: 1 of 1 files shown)


norkans7:
That is intentional as the first story is shown on the home page on https://github.com/rapidpro/ureport/blob/master/templates/public/index.haml#L133

Just make sure you are running the latest changes a recent fix was added in https://github.com/rapidpro/ureport/pull/658

PR #676, opened 2021-02-17
Use poetry

norkans7:


ureport/utils/tests.py
```diff
@@ -42,6 +42,19 @@
 )
 
 
+class MockResponse(object):
```
rowanseymour, line 45: This is defined in dash already right? Maybe reuse that one?

norkans7: I kept getting module not found since that class is defined in dash_test_runner that is not being recognized.

Probably we can more that class to https://github.com/rapidpro/dash/blob/master/dash/test.py ?

rowanseymour: Ah yeah since `dash_test_runner` isn't included in the final dependency - yeah move it to test.py

code_check.py
```diff
@@ -84,6 +84,8 @@ def get_current_msgids():
     update_po_files()
     status("Recompiling locale MO files")
     cmd("python manage.py compilemessages")
+    status("Exporting requirements")
+    cmd("poetry export -f requirements.txt -o pip-freeze.txt --without-hashes")
```
norkans7, line 88: Make sure pip-freeze is up to date as we still rely on that for deploys

PR #678, opened 2021-02-22
Add Macedonia's flag

rafaelbreno:


norkans7:
@rafaelbreno the country icon is not correct. Reverted the PR

U-Report U-Report Global 2021-03-01 15-16-28

PR #712, opened 2021-05-11
Added São Tomé and Príncipe flag

Sandro-Meireles:


ureport/settings_common.py
```diff
@@ -1405,6 +1405,13 @@ def traces_sampler(sampling_context):  # pragma: no cover
         countries_codes=["ROU"],
         count_link="http://romania.ureport.in/count/",
     ),
+    dict(
+        name="São Tomé and Príncipe",
+        host="//stp.ureport.in",
+        flag="flag_stp.png",
+        countries_codes=["STP", "ST"],
```
norkans7, line 1412: We only use alpha 3 codes and each country is added once, please remove "ST"

norkans7 (changes requested):
remove ST code

ureport/settings_common.py
Sandro-Meireles: Okay, I uploaded a commit with this fix.

norkans7 (approved):

PR #714, opened 2021-05-17
Add Kenya flag

Sandro-Meireles:

PR #715, opened 2021-05-17
Add Italia flag

Sandro-Meireles:


norkans7 (changes requested):
Incorrect size

norkans7:
@rafaelbreno can you fix this so we push that the servers at the same time as #714 ?

norkans7 (approved):

PR #724, opened 2021-05-26
Reduce query numbers

norkans7:


norkans7:
replaced by #726

norkans7:
we already run the task once a day

PR #740, opened 2021-07-15
Flow results 1

norkans7:
This has the first step to add models that will be storing data that do not change/updated by the admin


norkans7:
contains #742 and #743

ureport/flows/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class FlowResult(models.Model):
+    is_active = models.BooleanField(default=True)
+
+    org = models.ForeignKey(Org, on_delete=models.PROTECT)
+
+    flow_uuid = models.CharField(max_length=36)
+
+    result_uuid = models.CharField(max_length=36)
```
nicpottier, line 15: Why not UUID fields? They take up far less space in the DB.

ureport/flows/models.py
```diff
@@ -0,0 +1,53 @@
...
+
+    flow_result = models.ForeignKey(FlowResult, on_delete=models.PROTECT, related_name="result_categories")
+
+    category = models.TextField(null=True)
+
+    @classmethod
+    def update_or_create(cls, flow_result, category):
+        existing = cls.objects.filter(flow_result=flow_result, category=category)
+
+        if existing:
```
nicpottier, line 48: `existing.exists()` otherwise you are doing a full select and throwing away the result. Probably really best to just assign `.first()` to obj here and update is_active and call save with `update_fields` as then you will be returning a category that has `is_active` set to True as it should.

nicpottier (approved):
Some tweaks but looks ok.

ureport/flows/models.py
norkans7: FLOIP backend does not use valid UUIDs

nicpottier (approved):

PR #749, opened 2021-07-27
Dedupe poll response categories

norkans7:


nicpottier:
`PollResponseCategory` just maps to actual categories on RapidPro flows ya? Any reason we aren't just calling the API and going from there?

norkans7:
We currently allow admin to customize how the category will be displayed as.
The poll is added we call the RapidPro flows API to get the definition and parse that to create the possible questions and categories for each.

norkans7:
@nicpottier something else I am missing?

nicpottier:
Sorry, what is this deduping then?

norkans7:
question response categories, they use to have a rule_uuid and on RapidPro the category could be added twice such as "Other" now we want those to be unique and we no longer have the field rule_uuid from RapidPro so we will remove that as well

nicpottier:
Ok ok.

nicpottier (approved):

PR #752, opened 2021-08-03
Add Madagascar flag

evansmurithi:


norkans7:
@evansmurithi please share a screenshot of the URL /icons_display/ to see the icon has the correct dimensions

evansmurithi:
@norkans7 unfortunately the version running on madagascar.ureport.in doesn't have that url

norkans7:
You can have the screenshot form the dev environment

evansmurithi:
@norkans7 here's the screenshot with the highlighted country flag
ureport-rivers-onalabs-org-icons_display-

norkans7:
I think the icon is does not have the correct size there and also it should be sorted alphabetically by country name.
You should not highlight as this is the test to make sure the file has the proper size

You probably need to reach out to the CO or the design team for the correct icon to use there as you see all the others have the same size

evansmurithi:
@norkans7 noted

evansmurithi:
@norkans7 find updated screenshot below:
ureport-rivers-onalabs-org-icons_display-

norkans7:
Looks goods @evansmurithi

PR #754, opened 2021-08-12
Update to django 3.2

norkans7:


ureport/settings_common.py
```diff
@@ -100,7 +102,7 @@ def traces_sampler(sampling_context):  # pragma: no cover
     ("uk", "Ukrainian"),
     ("uz", "Uzbek"),
     ("my", "Burmese"),
-    ("mk_MK", "Macedonian"),
+    ("mk-mk", "Macedonian"),
```
norkans7, line 105: @matmsa27 @johncordeiro Macedonian org needs to adjust the language to this, the previous was wrong

PR #756, opened 2021-08-16
Fix bugs in contacts triggers

norkans7:


ureport/sql/contacts_0025.sql
```diff
@@ -0,0 +1,167 @@
...
+DECLARE
+  _count INT;
+BEGIN
+  IF _add THEN
+    _count = 1;
+  ELSE
+    _count = -1;
+  END IF;
+  -- If we have a org, increment all reporters counters for its values
+  IF _contact.org_id IS NOT NULL THEN
```
norkans7, line 25: the only change is this line from 

```IF _contact.org_id IS NOT NULL AND _contact.is_active THEN```

File: ureport/contacts/tests.py [hunk 2]
```diff
@@ -214,6 +214,9 @@ def test_recalculate_reporters_stats(self):
             state="R-LAGOS",
             district="R-OYO",
         )
+        contact.is_active = False
+        contact.save()
+
```
norkans7, line 219: I adjusted the test to be more realistic on how we end up in that state,
I think there is no way to create a contact with is_active=False without first being is_active=True and update that when the contact is removed from the reporter group.

ureport/sql/contacts_0025.sql
ericnewcomer: Are you sure that is enough?

I see below where we are looking at registered_on to determine if gender, born, etc should be updated. Do contacts still have a registered_on if they were considered active and then later removed from the group on RP?

norkans7: The counts `registered_gender`, `registered_born`... are for the engagements page sign up rate.

So a contact that is removed from the reporters group still has the fields set so we are decrementing the counts for back when they signed up.

Probably we should not decrement the time they signed up but also keeping that would allow the same contact to mess the stats just by joining and leaving multiple times. So I chose to always decrement.

ureport/contacts/tasks.py
```diff
@@ -21,9 +22,12 @@
 
 @app.task(name="contacts.rebuild_contacts_counts")
 def rebuild_contacts_counts():
+    r = get_redis_connection()
     orgs = Org.objects.filter(is_active=True)
     for org in orgs:
-        Contact.recalculate_reporters_stats(org)
+        key = TaskState.get_lock_key(org, "contact-pull")
```
norkans7, line 28: We grab the lock so when this is happening we are sure no task is fetching more contacts.

PR #766, opened 2021-08-31
Add tags

norkans7:


templates/public_base.haml
```diff
@@ -407,6 +407,11 @@
                 %a.admin-menu.text-sm.leading-loose(href="{% url 'categories.category_list' %}")
                   - trans "Categories"
   
+            - if request.user.is_superuser or org_perms.tags.tag_list
+              %li
+                %a.admin-menu.text-sm.leading-loose(href="{% url 'tags.tag_list' %}")
+                  - trans "Tags"
+
```
ericnewcomer, line 414: Can we get a screenshot of what this looks like? I presume this is so admins can cleanup tags across the site? They can delete them and we'll just remove them from the polls?

norkans7:
Homepage and admin toolbar on top
U-Report Nigeria 2021-08-31 23-32-31

Tags list and tags link on the left menu
U-Report 2021-08-31 23-32-54

Create tag
U-Report 2021-08-31 23-33-15

Edit tag
U-Report 2021-08-31 23-33-34

Add tags to poll
U-Report 2021-08-31 23-34-14

Pol tags display
U-Report Nigeria 2021-08-31 23-35-11

ericnewcomer:
Nice this helps. Another question, is the role the same for adding tags to a poll the same as creating a tag? If so, they really should be able to add the tag inline via the select2 widget, which doesn't look possible here.

norkans7:
I added a way to add new tag inline

PR #941, opened 2022-02-10
Sec updates

norkans7:


norkans7:
Should move first to tailwind CSS v3 and support RTL with that

PR #949, opened 2022-03-17
Support to "el" (greek) language

matmsa27:


norkans7:
I think we need to include the .mo file too

matmsa27:
Oh, sorry @norkans7 , thanks for reminding. I have added the `.mo` file for el language.

norkans7 (approved):

PR #1077, opened 2022-08-12
Add Mauritania flag

norkans7:


norkans7:
Monosnap nigeria ureport io:8000:icons_display: 2022-08-12 11-53-56

PR #1157, opened 2023-09-27
Add Slovenija Flag

brunocavalcantelima:
We request the addition of the flag of Greece, Here's the example:
eg


norkans7:
I guess you mean Slovenia! @brunocavalcantelima

brunocavalcantelima:
Hi, Norkans

The client preferred the translation of the country's name to be slovenian and not in English language.

PR #1163, opened 2023-10-16
Update translations locale for en lc

elitonzky:


matmsa27:
Hi @norkans7 we uploaded the branch, but some problems with codecove token happened

norkans7:
Thank you, I will merge that.

PR #1360, opened 2026-07-27
Add org object permission check to poll pull refresh view

norkans7:
PullRefresh was the only poll CRUDL action without an org object permission check and had an unfiltered queryset, so any authenticated user holding the `polls.poll_pull_refresh` permission could trigger a pull refresh on another org's poll by pk.

Since pull refresh deletes and re-pulls all results for a poll, it is now restricted to superusers and staff users only — org administrators and editors cannot trigger it.

Tests cover: unauthenticated denial, org admin and editor denial, staff and superuser success.

### DIFF
--- a/ureport/polls/tests.py
+++ b/ureport/polls/tests.py
@@ -94,13 +94,37 @@ class PollTest(UreportTest):
             response = self.client.post(pull_refresh_url, post_data, SERVER_NAME="uganda.ureport.io")
             self.assertLoginRedirect(response)
 
-            self.login(self.admin)
+            # an admin of a different org cannot pull refresh this poll, from their own site
+            other_org_admin = self.create_user("OtherOrgAdmin")
+            self.nigeria.administrators.add(other_org_admin)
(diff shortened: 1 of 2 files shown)


ureport/polls/tests.py
```diff
@@ -94,13 +94,37 @@ def test_poll_pull_refresh(self):
...
+            response = self.client.post(pull_refresh_url, post_data, SERVER_NAME="nigeria.ureport.io")
             self.assertLoginRedirect(response)
+            mock_pull_refresh.assert_not_called()
 
+            # nor from the poll org's site, where they are not a member
             response = self.client.post(pull_refresh_url, post_data, SERVER_NAME="uganda.ureport.io")
             self.assertLoginRedirect(response)
+            mock_pull_refresh.assert_not_called()
+
+            # an admin of the poll's own org can pull refresh it
```
norkans7: Adjust so only superuser and staff are allow to pull to access to pull refresh

PR #1363, opened 2026-07-27
Restrict custom_html org config to superusers

norkans7:
The `custom_html` org config value is rendered on public pages inside an `{% autoescape off %}` block — it is the only org config field output raw — so whoever can edit it can inject arbitrary script into every visitor's browser. Unlike the other raw-output config fields it was not marked `superuser_only`, leaving it editable by any org administrator.

This gates it behind superuser like its siblings. Existing stored values continue to render on public pages, and an admin saving the org edit form does not clear them. Org admins keep the dedicated, properly-escaped analytics fields (Google Tag Manager, Google Analytics, Facebook Pixel) for the advertised use case.

### DIFF
--- a/ureport/settings_common.py
+++ b/ureport/settings_common.py
@@ -653,6 +653,7 @@ ORG_CONFIG_FIELDS = [
             required=False,
             widget=Textarea,
         ),
+        superuser_only=True,
     ),
     dict(
         name="other_languages_sites",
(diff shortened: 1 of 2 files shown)


ureport/settings_common.py
```diff
@@ -653,6 +653,7 @@
             required=False,
             widget=Textarea,
         ),
+        superuser_only=True,
```
norkans7, line 656: No, we want to allow admins to use that field

norkans7:
Closing — keeping custom_html editable by org admins is the intended behavior, so this change is not wanted.

PR #1393, opened 2026-09-15
Replace smartmin login with django-allauth

norkans7:
First of three PRs bringing django-allauth to U-Report (this one: account login; then MFA; then SSO providers).

## What changes

- **Login, logout, password change and password reset** are now allauth's account views under `/accounts/`, rendered inside the admin chrome through overrides of allauth's layout and element templates. Login is by email address. The old `/users/login/` and `/users/logout/` URLs redirect to the new ones.
- **New `ureport.users` app** (label `ureport_users`, since `smartmin.users` keeps the `users` label):
  - `AccountAdapter` closes self-service signup and only allows post-login redirects to relative paths or hosts under `HOSTNAME`. allauth's default honours the `*` wildcard in `ALLOWED_HOSTS`, which would have made `next` an open redirect.
  - Nobody is pre-verified: allauth records and confirms a user's email the first time they log in, so existing users get a confirmation email on their next login. A `post_save` receiver lowercases emails (allauth stores and matches them in lowercase) and removes address records a user no longer has when their email changes.
  - The staff user CRUDL keeps list, create, update and profile under the existing `users.user_*` names; the "log in as" action is gone. The profile page now edits name only and links to allauth's password and email pages.
- `SITE_ALLOW_NO_ORG` lists every allauth account URL name so the pages work on the root host, guarded by a test that walks `allauth.urls`.
- `settings.py.docker` gains an optional `SESSION_COOKIE_DOMAIN` env var (also applied to the CSRF cookie) so a session started on one subdomain is valid on all of them, which the SSO PR will rely on.
- smartmin's failed-login lockout is replaced by allauth's cache-backed rate limits.
- `.gitignore` entries for the `dash`/`smartmin` dev symlinks are anchored to the repo root; the unanchored form was hiding new files under `templates/smartmin/`.

Most of the diff is the locale files: the new strings triggered a regeneration, and current gettext rewraps multi-line msgids even with `--no-wrap`. Only these msgids actually changed: added "Hello,", "The U-Report team", "Sign In", "Change Password", "Change Email"; removed "Error".

## Operator notes

- Set `SESSION_COOKIE_DOMAIN` to `.<SITE_HOSTNAME>` when running multiple org subdomains.
- Accounts with a blank email cannot log in under email login. Where two accounts share an email, the first to confirm it keeps it; allauth refuses to confirm an address another account has already verified.
- Existing bookmarks to `/users/login/` keep working via a redirect.

## Testing

Full suite passes (421 tests). New tests cover login by email, the first-login email confirmation flow, safe/unsafe `next` handling, logout, legacy URL redirects, closed signup, password reset email, the org middleware allowlist, profile and email address handling.


ureport/users/views.py
```diff
@@ -0,0 +1,42 @@
...
+        field_config = {
+            "username": dict(readonly=True, label=_("Username")),
+            "email": dict(readonly=True, label=_("Email")),
+        }
+        template_name = "smartmin/users/user_profile.html"
+
+        def has_permission(self, request, *args, **kwargs):
+            return self.request.user.is_authenticated
+
+    class Mimic(SmartUserCRUDL.Mimic):
```
norkans7: Remove the mimic view

ureport/users/migrations/0001_backfill_email_addresses.py
```diff
@@ -0,0 +1,41 @@
...
+        .exclude(id__in=users_with_address)
+        .order_by(F("last_login").desc(nulls_last=True), "id")
+    )
+    for user in users:
+        email = user.email.strip().lower()
+        if email in taken:
+            logger.warning("skipping user %s (%s) as that email already belongs to another account", user.pk, email)
+            continue
+
+        EmailAddress.objects.create(user=user, email=email, verified=True, primary=True)
```
norkans7: Existing users should not be verfied, they need to be verified on the next login

